PR #19, opened 2023-10-17
[feat] create text input component

ccatherinetan:

* Create input text component: Input Title + Input Box 
* Added more colors to `styles/colors.ts`: 
  * inputSuggestionColor: '#B3B3B3', // light grey 
  * existingInputColor: '#555555', // dark grey 
  * errorColor: '#D13E40', // red
  * activeGreen: '#097A62'

Inactive, No Input
inactive_noinput
Active, No Input
active_noinput
Active, Existing Input
active_existinginput
Inactive, Existing Input
inactive_existinginput
Error, No Input (for 
Screen Shot 2023-10-19 at 4 03 19 PM
both active and inactive)
Error, Existing Input  (for both active and inactive)
Screen Shot 2023-10-19 at 4 07 47 PM

* border is 2px

See both files in `components/TextInput`
* `components/TextInput/styles.tsx`: styling the Input Title and the Input Box 
* `components/TextInput/TextInput.tsx`

Run `npm run dev`, navigate to http://localhost:3000/input-text
* check `app/input-text/page.tsx`
* can change the label, placeholder, erroring, and errorText 
* type in an input, check that it appears in a dark grey 
* check that focus works (i.e. when you click on the text box, the green border appears)
* check that error works: change the value of `$error` to `true` in `components/TextInput/TextInput.tsx` (See the images for the expected behavior of errors) 

## :seedling: Next steps 

CC: @pragyakallanagoudar


package.json
```diff
@@ -20,13 +20,14 @@
     "dotenv": "^16.3.1",
     "eslint-config-next": "13.5.2",
     "eslint-import-resolver-typescript": "^3.6.1",
-    "next": "13.5.2",
+    "next": "13.5.4",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "styled-components": "^6.0.8"
+    "styled-components": "^6.0.9"
```
jinkang-0: Looks like you might have ran `npm i` instead of `npm ci`, so it updated some packages.

@pragyakallanagoudar do we want to keep this change to update the dependencies? It could be easier to do it now than later (to avoid hairy debugging).

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { InputTitleText, InputText } from './styles';
+
+export default function TextInput() {
+  return (
+    <>
+      <InputTitleText>Label</InputTitleText>
+      <InputText $error={false} placeholder="Input Suggestion" />
+    </>
```
jinkang-0: Instead of making a test component containing these components, then inserting this into a test page, I think it would be cleaner to just put these components into the test page itself.

Unless we want to keep this component? Personally I don't think it's the best idea since it's just bundling two components together, which would be more clunky to use.

src/components/TextInput/styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+    $error ? COLORS.errorColor : COLORS.existingInputColor};
+
+  &:focus {
+    outline: none !important;
+    border: 1px solid
+      ${({ $error }) => ($error ? COLORS.errorColor : COLORS.activeGreen)};
+  }
+  &::placeholder {
+    color: ${({ $error }) =>
+      $error ? COLORS.errorColor : COLORS.inputSuggestionColor};
```
jinkang-0: We may not want the placeholder color to be red (or at least the same red as the value color), since it will get confusing to the user whether or not the text in the input field is there. It's also a lot of red color on an otherwise neutral background (white).

If we want to convey to the user that the error was because of no value, we could insert a small text below the input (handled outside of the component), and the red border can also convey that something is wrong here.

src/components/TextInput/styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+`;
+export const InputTitleText = styled.label`
+  color: #555555;
```
jinkang-0: Should change to use the value from `COLORS` so it stays consistent should we ever edit the `COLORS` table.

src/components/TextInput/styles.tsx
```diff
@@ -0,0 +1,41 @@
+import styled from 'styled-components';
+import { Inter } from 'next/font/google';
+import COLORS from '../../styles/colors';
+
+const inter = Inter({ subsets: ['latin'] });
+
+export const InputText = styled.input<{ $error: boolean }>`
+  color: ${COLORS.existingInputColor};
+  font-size: 1rem; // 16px
+  font-family: ${inter.style};
```
jinkang-0: `inter.style` by itself looks like `font-family: ...` already, no need to add `font-family:` before it.

src/components/TextInput/styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+  font-weight: 400;
+  line-height: 19.2px;
+  word-wrap: break-word;
+  padding: 10px 20px 10px 20px;
```
jinkang-0: I would recommend changing to use `rem` units, since the padding is dependent on the font size. If the user decides to adjust their browser font size, the padding would look off.

src/components/TextInput/styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+  &::placeholder {
+    color: ${({ $error }) =>
+      $error ? COLORS.errorColor : COLORS.inputSuggestionColor};
+  }
+`;
+export const InputTitleText = styled.label`
+  color: #555555;
+  font-size: 1.5rem; // 24px;
+  font-family: ${inter.style};
```
jinkang-0: flagging this here for the `font-family` thing as well.

jinkang-0 (changes requested):

src/components/TextInput/styles.tsx
jinkang-0: Also for line-height

package.json
jinkang-0: Though also it seems to be minor enough updates that we can worry about that for later

src/components/TextInput/styles.tsx
jinkang-0: Also could just do `padding: 0.625rem 1.25rem;`, which shortcuts to the equivalent form. The first parameter is vertical padding, the second is horizontal.

package.json
pragyakallanagoudar: think we should do it now! ty for flagging, we can hold onto this change

jinkang-0:
I believe from the design on Figma that when it is inactive, the border should be gray instead of green?

Figma:
image

Current:
image

package.json
jinkang-0: On second thought this is likely due to `npm audit fix --force`, since no changes happened to the `package.json` when I tried to run `npm i`. Still likely good to do a quick package update.

package.json
```diff
@@ -20,13 +20,14 @@
...
   },
   "devDependencies": {
     "@calblueprint/eslint-config-react": "^0.0.3",
+    "@types/styled-components": "^5.1.28",
```
jinkang-0: Curious: do you remember what this was for? Do we need this for anything?

src/components/TextInput/TextInput.tsx
ccatherinetan: sounds good! i'll just put the two components separately on the test page.

src/components/TextInput/styles.tsx
ccatherinetan: mmm i get what you're saying! On the figma, even the placeholder text is red, so i'm going to loop in @davidqing6432 for his thoughts!

src/components/TextInput/styles.tsx
ccatherinetan: got it!

src/components/TextInput/styles.tsx
ccatherinetan: changed! thank you!

package.json
ccatherinetan: My import of styled components was erroring, and it told me to add this line! however, the red error line still appears, even with this line included :<

`import styled from 'styled-components';` has the red squiggles. The error reads: 

`'@types/styled-components' should be listed in the project's dependencies. Run 'npm i -S @types/styled-components' to add iteslint[import/no-extraneous-dependencies](https://github.com/import-js/eslint-plugin-import/blob/v2.28.1/docs/rules/no-extraneous-dependencies.md)`

src/components/TextInput/styles.tsx
ccatherinetan: I'm a little confused about what you means but i'll look into it!

jinkang-0: I meant that if you just inserted `${inter.style}` like that without `font-family: ` before it, it would still work fine, since `inter.style` already includes the property name.

src/components/TextInput/styles.tsx
jinkang-0: yeah definitely a david type question 🖌️

package.json
jinkang-0: Weird... let's look into it tomorrow

src/app/input-text/page.tsx
```diff
@@ -1,7 +1,13 @@
...
+import { InputTitleText, InputText } from '../../components/TextInput/styles'; 
 
-export default function InputText() {
-  return <TextInput />;
-}
+export default function InputTextTest() {
+  return (
+    <>
+      <InputTitleText>Label</InputTitleText>
+      <InputText $error={false} placeholder="Input Suggestion"/>
```
pragyakallanagoudar: Thanks for making this change Catherine! I meant to loop David into this before but forgot— @davidqing6432 can you tell us if you anticipated the label and text input being a single component, like Catherine had before? In the Figma it seemed like this was what was expected, so I just wanted to clarify. 

It seems like the text input never occurs without the label, so it might make sense to make them one component for that reason? Although Jinkang brings up a good point that this allows for more flexibility and is simpler.

ccatherinetan:
> I believe from the design on Figma that when it is inactive, the border should be gray instead of green?
> 
> Figma: image
> 
> Current: image

fixed! nice catch!

src/components/TextInput/styles.tsx
ccatherinetan: i incorporated david's feedback! see updated screenshots

src/app/input-text/page.tsx
ccatherinetan: implemented the combined component as we discussed :)

package.json
```diff
@@ -27,6 +27,7 @@
   },
   "devDependencies": {
     "@calblueprint/eslint-config-react": "^0.0.3",
+    "@types/styled-components": "^5.1.29",
```
jinkang-0, line 30: Do we still need the `@types/styled-components` package? Not sure if the error showed up again..

src/components/TextInput/styles.tsx
```diff
@@ -0,0 +1,55 @@
...
+  align-items: flex-start;
+  gap: 10px;
+  color: ${COLORS.existingInputColor};
+
+  &:focus {
+    outline: none !important;
+    border: 2px solid
+      ${({ $error }) => ($error ? COLORS.errorColor : COLORS.activeGreen)};
+  }
+  &::placeholder {
```
jinkang-0, line 30: VERY MINOR: maybe an extra line break before this line 🍵

package.json
pragyakallanagoudar: When she got rid of it, the PR failed the TypeScript compiler Github action :(( seems like we might need it!

jinkang-0: rip I'll sort thru the eslint package stuff

jinkang-0 (approved):

src/components/TextInput/styles.tsx
ccatherinetan: will do!

PR #26, opened 2023-10-26
[feat] create full size button for onboarding and auth

rahihazra:

- Creates a full size button to be used for the onboarding and auth pages

- there is a customized button component (that will be changed in the future due to design)
- since we had no place to put the button, i created a test page to display only the component

Screen Shot 2023-10-26 at 1 30 02 PM

- order in which to review: src/components/BigButton.tsx and then src/app/button-test/page.tsx
- we don't have a specific place to implement this button, so changes will need to be made

## :seedling: Next steps 

CC: @pragyakallanagoudar


src/components/BigButton.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
+
+interface BigButtonProps {
+  color?: string;
```
jinkang-0: I think the convention for styled component props is to have a `$` prefix attached to them, so like `$color`, which is just to differentiate them from React components I think? Not too sure what conventions we're following though.

src/components/BigButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+  onChange?: () => void;
```
jinkang-0: We don't need to add an `onChange` prop here, since the `BigButton` builds off of a regular HTML button (signified by `styled.button`), it already has `onChange` as a possible prop, so we don't need to (and can't, in styled components) handle that!

In the future, it's useful to keep in mind that props to a styled component is meant to be help decide how to style. In the case of `onChange`, it won't help us decide how to style this button (i.e. what color, what padding, etc etc)

src/components/BigButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+const BigButton = styled.button<BigButtonProps>`
+  background-color: ${props => props.color || 'blue'};
+  color: white;
+  padding: 10px 20px;
```
jinkang-0: Ideally it's good practice to use `rem` units instead of `px`! By default, most browsers have it set such that `1rem` = `16px`, so we can do unit conversion to adjust that. The reason why is so things scale nicely with `font-size`.

src/components/FullSizeButton.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import styled from 'styled-components';
+
+interface BigButtonProps {
+  color: string;
+  onChange: () => void;
+}
+
```
jinkang-0: Not sure what this file is used for...? Can we remove this file?

src/app/button-test/page.tsx
```diff
@@ -0,0 +1,8 @@
+'use client';
+
+import React from 'react';
+import BigButton from '@/components/BigButton';
+
+export default function ButtonTest() {
+  return <BigButton>Big Button</BigButton>;
+}
```
jinkang-0: Nice! I think we should remove testing pages before merging them in ideally though, but maybe you can comment this out and move it into `BigButton.tsx` to show how to use it?

src/components/BigButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+  border: none;
+  border-radius: 5px;
```
jinkang-0: Flagging this here as well for `rem` unit conversion! `border-radius` is one of those things that can get finicky, but I think we do want it to scale with `font-size`.

jinkang-0 (commented):
Looks good so far! I left some comments for improvements, but it should be good to merge after they're addressed!

src/components/BigButton.tsx
pragyakallanagoudar: Let's change to this! Thanks Jinkang

src/components/BigButton.tsx
pragyakallanagoudar: Agree! Also, since we're only passing in the one color prop, I think we can go ahead and get rid of this interface. Instead, your styled component would look something like:

```
const BigButton = styled.button<{ $color?: string }>`
// ...etc.
```

src/app/button-test/page.tsx
pragyakallanagoudar: +1! let's keep this here for right now because we're not ready to merge yet. once we are, we can do what Jinkang recommended

pragyakallanagoudar: Also, let's pass in some props to the button so we can test it:

```
<BigButton color={COLORS.abaBlue} onChange={() => console.log('button clicked')}>
      Big Button
</BigButton>
```

src/components/BigButton.tsx
pragyakallanagoudar: https://nekocalc.com/px-to-rem-converter

jinkang-0 (approved):
lgtm!

pragyakallanagoudar (approved):
Looks great! For reference, we commented out a bunch of things to prevent breaking from schema changes, and also commented out the button test screen and moved it to the Button file.

PR #29, opened 2023-10-30
[refactor] update schema types (again)

jinkang-0:

Modified `src/types/schema.ts`:
- Updated existing types to reflect schema changes on drawSQL
- Added miscellaneous types for table fields for specific `enums` and join tables

Updated the following pages to handle new type changes:
- `src/app/cases/page.tsx`
- `src/app/create-profile/page.tsx`
- `src/app/interest/page.tsx`
- `src/app/profile/page.tsx`

Updated the following components to handle new type changes:
- `src/components/ListingCard.tsx`

Unplanned:
- Modified queries in `src/api/supabase/queries/` to have consistent error messages.
- Updated comments in `src/utils/helpers.ts`

First review changes in `types/schema.ts`, then review in any order.

## :seedling: Next steps 

CC: @pragyakallanagoudar


src/types/schema.ts
```diff
@@ -1,65 +1,108 @@
...
-  is_remote: boolean;
-  in_court: boolean;
-  needs_attorney: boolean;
-  needs_interpreter: boolean;
-  upcoming_date: string;
-  relief_sought: string;
-  active: boolean;
+// table field types
+export type ImmigrationLawExperienceEnum =
+  | 'No Experience'
```
pragyakallanagoudar: Can we change this to `LOW`, `MEDIUM`, `HIGH`? Sorry for the absurd number of changes to this, but David and I ultimately decided that these words would represent no prior cases, 1 prior case, and multiple prior cases, respectively. See this from interest form design:

Screenshot 2023-10-31 at 8 41 45 PM

src/types/schema.ts
```diff
@@ -1,65 +1,108 @@
...
+  needs_interpreter?: boolean;
+  upcoming_date?: string;
+  experience_needed: ImmigrationLawExperienceEnum;
 }
 
 export interface Profile {
   user_id: UUID;
   first_name: string;
   last_name: string;
+  preferred_first_name?: string;
```
pragyakallanagoudar: think we're going to cut this after all!

src/types/schema.ts
```diff
@@ -1,65 +1,108 @@
...
   };
 }
+
+// supabase: auth -> users table row
+export interface AuthUser {
+  id: UUID;
+  email: string;
+}
+
+// irrelevant until mvp
```
pragyakallanagoudar, line 82: LOL thank you for including these!

src/types/schema.ts
```diff
@@ -1,65 +1,108 @@
...
   form_response: {
-    interestType: string;
+    rolesInterested: RoleEnum[];
     interestReason: string;
     start_date: string;
   };
 }
+
+// supabase: auth -> users table row
+export interface AuthUser {
```
pragyakallanagoudar: Curious why we need this if we're going to just keep getting the `userId` from the session as implemented in the `ProfileContext`? i.e. are you thinking of adding an `AuthUser` object to the `ProfileContext`? (That would be good I think!)

src/api/supabase/queries/cases.ts
```diff
@@ -3,18 +3,18 @@ import supabase from '../createClient';
 
 export async function getAllCases() {
   const { data, error } = await supabase.from('cases').select();
-  if (error) throw error;
+  if (error) throw new Error(`Error fetching all cases: ${error.message}`);
```
pragyakallanagoudar, line 6: yay thank you for more specific error messages!

src/types/schema.ts
```diff
@@ -1,65 +1,108 @@
...
+  | 'Research Fellow'
+  | 'Translator';
+
+export type ListingTypeEnum =
+  | 'Case'
+  | 'Limited Assistance'
+  | 'Translation Request';
+
+// join table rows
+export interface UserLanguage {
```
pragyakallanagoudar: Curious to hear your thought process on naming this `UserLanguage` as opposed to `ProfileLanguage`? I think if we do `UserLanguage` here I'm almost inclined to rename `Profile` to `UserProfile`....

src/types/schema.ts
jinkang-0: I think I saw it already there on the file and didn't question it... Likely not needed though so I'll get rid of it.

src/types/schema.ts
jinkang-0: I think my thought process was that naming it `ProfileLanguage` means naming the other one `ProfileRole` which sounds similar to `ProfileRow` and I got confused. I'll change it to `Profile-` to match the schema.

src/types/schema.ts
jinkang-0: Do we want one-word names for the other `enum`s as well?

src/types/schema.ts
```diff
@@ -1,65 +1,105 @@
...
-  relief_sought: string;
-  active: boolean;
+// table field types
+export type ImmigrationLawExperienceEnum = 'LOW' | 'MEDIUM' | 'HIGH';
+
+export type AdjudicatingAgencyEnum =
+  | '9TH_CIRCUIT'
+  | 'BIA'
+  | 'IMMIGRATION_COURT'
+  | 'IMMIGRATION_ENFORCEMENT'
```
pragyakallanagoudar: Can we rename this to `ICE`?

src/types/schema.ts
varortz: thank u jinkang for taking care of the schema changes!!

src/types/schema.ts
```diff
@@ -1,65 +1,105 @@
...
+// sql table rows
+export interface CaseListing {
   id: UUID;
-  summary: string;
-  languages: string[];
-  interest_ids: UUID[];
+  title?: string;
+  summary?: string;
+  country?: string;
+  client_location?: string;
```
varortz, line 46: I think these four fields weren't optional before, wondering what the reason was for making them optional now? Does this have something to do with legalserver?

src/types/schema.ts
```diff
@@ -1,65 +1,105 @@
...
+  legal_server_id: number;
+  hours_per_month?: number;
```
varortz, line 48: do we want `hours_per_month` to be optional? I think David added a tag on the case listing card component that corresponds to a case's anticipated hrs/month. Are we ok with not displaying that tag for all cards if some cases don't have that info, or should we make it required to show the hrs/month data for all case listing cards?

pragyakallanagoudar: great point!! the hours per month is an additional field that is going to be added to LegalServer, and so there is the possibility of it being `null` if someone forgets to input it on LegalServer. generally we're making all LegalServer fields nullable because there's always the possibility of someone forgetting to fill them out on there.

@davidqing6432 wondering if you have ideas for modifications to the case listing card in the case that some of these fields are null?

src/types/schema.ts
pragyakallanagoudar: addressed in your comment about `hours_per_month`!

src/types/schema.ts
```diff
@@ -1,65 +1,105 @@
...
-  in_court: boolean;
-  needs_attorney: boolean;
-  needs_interpreter: boolean;
-  upcoming_date: string;
-  relief_sought: string;
-  active: boolean;
+// table field types
+export type ImmigrationLawExperienceEnum = 'LOW' | 'MEDIUM' | 'HIGH';
+
+export type AdjudicatingAgencyEnum =
```
pragyakallanagoudar: Related to your message on Slack, what are your thoughts on making adjudicating agency a string instead of an enum? I think that might allow for more flexibility, in case they add another potential agency afterward

src/types/schema.ts
```diff
@@ -1,65 +1,105 @@
...
-  id: UUID;
   listing_id: UUID;
   user_id: UUID;
-  listing_type: number;
+  listing_type: ListingTypeEnum;
   form_response: {
-    interestType: string;
+    rolesInterested: RoleEnum[];
     interestReason: string;
     start_date: string;
```
pragyakallanagoudar: is it not possible to make `start_date` of type `timestamptz`?

src/components/InterestForm/index.tsx
```diff
@@ -42,12 +41,14 @@ export default function InterestForm({ caseData }: { caseData: CaseListing }) {
...
-        id: crypto.randomUUID() as UUID,
         listing_id: caseData.id,
-        listing_type: 1,
+        listing_type: 'CASE',
         user_id: '36b8f84d-df4e-4d49-b662-bcde71a8764f',
         form_response: {
-          interestType: rolesInterested,
+          rolesInterested:
+            rolesInterested === radioOptions[2]
+              ? ['ATTORNEY', 'INTERPRETER']
```
pragyakallanagoudar, line 50: NICE

src/app/cases/page.tsx
```diff
@@ -76,29 +76,20 @@ export default function Page() {
...
           multi
-          // better solution available if we update ts target to es6
-          options={['All countries'].concat(
-            caseData
-              .map(c => c.country)
-              .filter((v, i, arr) => arr.indexOf(v) === i),
-          )}
+          options={[
+            'All countries',
+            ...Array.from(
```
pragyakallanagoudar, line 81: 🤞 that the country of origin field is in a standardized format on LegalServer! also more generally, I suspect that the work of converting the LegalServer data into this form is going to be,,, curious,,, but we'll deal with that then

pragyakallanagoudar (approved):
Looks fantastic, Jinkang! I have a couple of clarifying questions and comments on here, but largely this looks good to merge. THANK YOU SO MUCH FOR WORKING ON THIS

src/types/schema.ts
jinkang-0: Yeah I think that makes sense given legal server data may not be standardized to fit into the enum

src/types/schema.ts
jinkang-0: I'll make it a js Date, thanks for flagging that!

src/app/cases/page.tsx
jinkang-0: 🤞

src/types/schema.ts
jinkang-0: We should also note that we will be pulling `experience_needed` from legal server as well, and we may want to ask them about the format it'll be in

PR #34, opened 2023-11-03
[feat] make progress bar component

jinkang-0:

Added `ProgressBar` component

Restructured auth flow to share layout with for general auth and for onboarding.

image

First, fetch this branch:
```
git fetch origin jinkang/ijp-69-make-progress-bar-component
```
Then, git checkout
```
git checkout jinkang/ijp-69-make-progress-bar-component
```
Finally, navigate to `localhost:3000/onboarding/` after starting the server to see the changes.

## :seedling: Next steps 

CC: @pragyakallanagoudar, @davidqing6432


src/components/ProgressBar/styles.tsx
```diff
@@ -0,0 +1,70 @@
+import COLORS from '@/styles/colors';
+import { H4 } from '@/styles/text';
+import styled from 'styled-components';
+
+export const ProgressBarContainer = styled.div`
```
pragyakallanagoudar: Can we change the border weight of the circles and the body to be a bit lighter? It feels slightly heavier / thicker than it is on the Figma

src/components/ProgressBar/styles.tsx
```diff
@@ -0,0 +1,70 @@
...
+export function ProgressCircle({
+  checked = false,
+  text,
+}: {
+  checked?: boolean;
+  text: string;
+}) {
+  return (
+    <ProgressCircleDiv $checked={checked}>
+      <H4>{text}</H4>
```
pragyakallanagoudar, line 65: Question for @davidqing6432: Personally, I feel like it would be better to have the titles for each circle be slightly further away from the circle itself?

src/app/(auth)/onboarding/layout.tsx
```diff
@@ -0,0 +1,19 @@
+'use client';
```
pragyakallanagoudar: Hmmm I know that this routing allows for us to have the onboarding `layout.tsx` inherit from the auth `layout.tsx`, but it feels like the routing may get a bit unwieldy if all the onboarding screens are in `src/app/(auth)/onboarding`, especially because onboarding isn't really part of auth. Thoughts?

src/app/(auth)/onboarding/layout.tsx
```diff
@@ -0,0 +1,19 @@
...
+export default function layout({ children }: { children: ReactNode }) {
+  // track progress with onboarding context
+  return (
+    <>
+      <ProgressBar
+        steps={new Set(['Basic Info', 'Languages', 'Legal Experience', 'Done'])}
+        progress={2}
+      />
+      {children}
+      <BigButton>Continue</BigButton>
```
pragyakallanagoudar: Can we make the text on this button white?

Also, curious how putting the button in the `layout.tsx` file would affect routing?

src/app/(auth)/onboarding/layout.tsx
jinkang-0: I think this is more of a question to discuss with all the devs, though it won't impact the user too much (will still be `/onboarding/...`). Personally I'm a fan of keeping it the way it is to keep consistency with Next.js' way of nesting layouts, since it looks pretty much the same. If we want to go for the `src/app/onboarding/` approach, we could, but I think to do so, we'll need a separate `AuthLayout` component that both uses, which doesn't feel as clean.

src/app/(auth)/onboarding/layout.tsx
jinkang-0: I believe the text color is a part of the Big Button sprint that is still in the backlog. The button placed here is because, on the Figma, there's a button at the bottom of all the onboarding pages that all said "Continue" (except for the last one 😔). We could remove it, up for discussion. Routing-wise, I'm thinking of handling the complexity in the onboarding `layout.tsx` and communicating with the onboarding pages using the onboarding context.

src/components/ProgressBar/styles.tsx
jinkang-0: Border weight and line (I'm assuming that's what you mean by the body) is the same as it is on Figma, 5px. I do think it looks off, but we should discuss that, because many auth screens also seem bigger than they should be. This may be a screen size issue, or a Figma issue.

pragyakallanagoudar (approved):

david-qing (approved):

PR #40, opened 2023-11-22
[feat] connect interest form to backend

varortz:

Reconnects interest form to backedn and implements stretch features for the interest form.

- Replaces text input for `startDate` with a date input
- Converts interest form's date input into a `DateInput` component 
- Adds error feedback to the `DateInput` and radio buttons on the interest form
- Implements displaying success message after submitting an interest
- Fixes small inconsistencies with padding, margins, etc
- Fixes colors on `TextInput` component

Screen Recording 2023-11-22 at 3 40 01 AM

- Checkout this branch & navigate to the `/cases` page
- **Form Submission:** Submit an interest using the form, go to the interests table on supabase, and make sure your interest was added to the table. You should see a success page after submitting an interest. **NOTE**: when checking for a submission on supabase, the number of interest records may not increase due to updating the insert query to an upsert query. I've noticed the submitted interests are usually on the first page of records at the very top, so might wanna check there first when testing!
- **Error Feedback:** Attempt to submit an interest form without the first two questions filled out. You should see error feedback for both of first questions. Make sure that the feedback disappears for each of the two questions after providing a response. Also attempt to submit a form with an invalid contact date (dates that aren't in the future, including today's date), you should see error feedback for this too and it should disappear after providing a valid date.

## :seedling: Next steps 
- Transform the Interest Form + Profile Match into a grid

## :link: Relevant Links
Cases Table (Supabase): https://supabase.com/dashboard/project/grffggxilkypzxwecaxz/editor/30417

N/A

CC: @pragyakallanagoudar

### DIFF
--- a/src/api/supabase/queries/interest.ts
+++ b/src/api/supabase/queries/interest.ts
@@ -1,7 +1,10 @@
 import { Interest } from '@/types/schema';
 import supabase from '../createClient';
 
-/** Get all interest from supabase interest table. */
+/**
+ * Fetches all interests from the database
+ * @returns a Promise of all interests objects 
+ */
 export async function getAllInterests() {
   const { data, error } = await supabase.from('interests').select();
 
@@ -11,7 +14,10 @@ export async function getAllInterests() {
   return data;
 }
 
-/** Insert a new interest object into supabase interest table. */
+/**
+ * Upserts an interest object into the database's interest table
+ * @param interest - an interest object 
+ */
 export async function upsertInterest(interest: Interest) {
   const { error } = await supabase.from('interests').upsert(interest).select();
   if (error) {
--- a/src/api/supabase/queries/interest.ts
+++ b/src/api/supabase/queries/interest.ts
@@ -12,8 +12,8 @@ export async function getAllInterests() {
 }
 
 /** Insert a new interest object into supabase interest table. */
-export async function insertInterest(interest: Interest) {
-  const { error } = await supabase.from('interests').insert(interest).select();
+export async function upsertInterest(interest: Interest) {
+  const { error } = await supabase.from('interests').upsert(interest).select();
   if (error) {
     throw new Error(`Error inserting an interest: ${error.message}`);
   }
--- a/src/api/supabase/queries/interest.ts
+++ b/src/api/supabase/queries/interest.ts
@@ -3,7 +3,7 @@ import supabase from '../createClient';
 
 /**
  * Fetches all interests from the database
- * @returns a Promise of all interests objects 
+ * @returns a Promise of all interests objects
  */
 export async function getAllInterests() {
   const { data, error } = await supabase.from('interests').select();
@@ -16,7 +16,7 @@ export async function getAllInterests() {
 
 /**
  * Upserts an interest object into the database's interest table
- * @param interest - an interest object 
+ * @param interest - an interest object
  */
 export async function upsertInterest(interest: Interest) {
   const { error } = await supabase.from('interests').upsert(interest).select();
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -10,7 +10,7 @@ import { openSans } from '@/styles/fonts';
       $secondaryColor (Required) - This color will determine the border color and background color on hover
    
     Example:
-      <Button $primarycolor={COLORS.blueMid} $secondarycolor={COLORS.blueDark} onClick={...}>
+      <Button $primaryColor={COLORS.blueMid} $secondaryColor={COLORS.blueDark} onClick={...}>
         [Button text here]
       </Button>
 
@@ -21,7 +21,7 @@ import { openSans } from '@/styles/fonts';
       $secondaryColor (Required) - This color will determine the border color and background color on hover
 
     Example: 
-      <Button $secondarycolor={COLORS.blueMid} onClick={...}>
+      <Button $secondaryColor={COLORS.blueMid} onClick={...}>
         [Button text here]
       </Button>
  */
--- a/src/components/DateInput/index.tsx
+++ b/src/components/DateInput/index.tsx
@@ -1,40 +1,35 @@
 import React, { Dispatch, SetStateAction } from 'react';
 import { InputLabel, ErrorText, InputDiv, InputDate } from './styles';
 
-type TextInputProps = {
+type DateInputProps = {
   label: string;
-  error: boolean;
-  errorText: string;
-  name: string;
+  error?: string;
+  name?: string;
   value: string;
-  filled: boolean;
   setValue: Dispatch<SetStateAction<string>>;
 };
 
 export default function DateInput({
   label,
-  error,
-  errorText = '',
+  error = '',
   name,
   value,
-  filled,
   setValue,
-}: TextInputProps) {
+}: DateInputProps) {
   return (
     <InputDiv>
-      {' '}
       <InputLabel>{label}</InputLabel>
       <InputDate
         required
         type="date"
         $error={error}
-        $filled={filled}
+        $filled={value !== ''}
         name={name}
         value={value}
         min={new Date().toISOString().split('T')[0]}
         onChange={e => setValue(e.target.value)}
       />
-      {error && <ErrorText>{errorText}</ErrorText>}
+      {error !== '' && <ErrorText>{error}</ErrorText>}
     </InputDiv>
   );
 }
--- a/src/components/DateInput/index.tsx
+++ b/src/components/DateInput/index.tsx
@@ -1,0 +1,40 @@
+import React, { Dispatch, SetStateAction } from 'react';
+import { InputLabel, ErrorText, InputDiv, InputDate } from './styles';
+
+type TextInputProps = {
+  label: string;
+  error: boolean;
+  errorText: string;
+  name: string;
+  value: string;
+  filled: boolean;
+  setValue: Dispatch<SetStateAction<string>>;
+};
+
+export default function DateInput({
+  label,
+  error,
+  errorText = '',
+  name,
+  value,
+  filled,
+  setValue,
+}: TextInputProps) {
+  return (
+    <InputDiv>
+      {' '}
+      <InputLabel>{label}</InputLabel>
+      <InputDate
+        required
+        type="date"
+        $error={error}
+        $filled={filled}
+        name={name}
+        value={value}
+        min={new Date().toISOString().split('T')[0]}
+        onChange={e => setValue(e.target.value)}
+      />
+      {error && <ErrorText>{errorText}</ErrorText>}
+    </InputDiv>
+  );
+}
--- a/src/components/DateInput/styles.tsx
+++ b/src/components/DateInput/styles.tsx
@@ -5,14 +5,9 @@ import { H4 } from '@/styles/text';
 
 const openSans = OpenSans({ subsets: ['latin'] });
 
-const getBorderColor = (error: boolean, filled: boolean): string => {
-  if (error) {
-    return COLORS.redMid;
-  }
-  if (filled) {
-    return COLORS.greyMid;
-  }
-  return COLORS.greyLight;
+const getBorderColor = (error: string, filled: boolean): string => {
+  if (error !== '') return COLORS.redMid;
+  return filled ? COLORS.greyMid : COLORS.greyLight;
 };
 
 export const InputLabel = styled(H4)`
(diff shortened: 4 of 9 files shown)


src/components/DateInput/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  error,
+  errorText = '',
+  name,
+  value,
+  filled,
+  setValue,
+}: TextInputProps) {
+  return (
+    <InputDiv>
+      {' '}
```
jinkang-0: do we need this extra space for anything? if not, can we get rid of it?

src/components/DateInput/index.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { Dispatch, SetStateAction } from 'react';
+import { InputLabel, ErrorText, InputDiv, InputDate } from './styles';
+
+type TextInputProps = {
+  label: string;
+  error: boolean;
+  errorText: string;
+  name: string;
+  value: string;
+  filled: boolean;
```
jinkang-0: it looks like we only use `filled` to determine whether or not the input has a value, but, can't we just determine that from the `value` prop?

src/components/DateInput/index.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { Dispatch, SetStateAction } from 'react';
+import { InputLabel, ErrorText, InputDiv, InputDate } from './styles';
+
+type TextInputProps = {
+  label: string;
+  error: boolean;
```
jinkang-0: similarly, I think we can determine whether or not there has been an error using the truthy value of `errorText`

src/components/DateInput/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  errorText: string;
```
jinkang-0: minor: we could leave `errorText` as an optional field and default it to `''` in the arguments field of the function component. Although there's no design that uses the `DateInput` without an error state, it's good practice to make it flexible for the future.

src/components/DateInput/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  name: string;
```
jinkang-0: `name` can also be optional, since this prop to the input is only used by HTML `form` elements on submit

src/components/DateInput/styles.tsx
```diff
@@ -0,0 +1,50 @@
+import styled from 'styled-components';
+import { Open_Sans as OpenSans } from 'next/font/google';
+import COLORS from '@/styles/colors';
+import { H4 } from '@/styles/text';
+
+const openSans = OpenSans({ subsets: ['latin'] });
```
jinkang-0: note: we may want to import the `openSans` instance from a `fonts.ts` file in the future

src/components/DateInput/styles.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const getBorderColor = (error: boolean, filled: boolean): string => {
+  if (error) {
+    return COLORS.redMid;
+  }
+  if (filled) {
+    return COLORS.greyMid;
+  }
+  return COLORS.greyLight;
```
jinkang-0: nit: we can turn these if statements to be one liners without the curly braces (i.e. `if (error) return COLORS.redMid`), could also consolidate the second if statement to a ternary statement (i.e. `return (filled) ? COLORS.greyMid : COLORS.greyLight`)

src/components/DateInput/styles.tsx
```diff
@@ -0,0 +1,50 @@
...
+`;
+
+export const InputDate = styled.input<{ $error: boolean; $filled: boolean }>`
+  ${openSans.style}
+  font-size: 1rem;
+  font-weight: 400;
+  color: ${props =>
+    props.$filled ? COLORS.greyDarker : COLORS.placeholderText};
+  border-radius: 0.313rem; // 5px
+  border: 2px solid ${({ $error, $filled }) => getBorderColor($error, $filled)};
```
jinkang-0: nit: can we stay consistent using either `props` or destructuring? (`props => ...` vs `({ $val }) => ...`)

src/components/InterestForm/styles.tsx
```diff
@@ -100,3 +98,10 @@ export const FormWarning = styled.p`
   max-width: 50%;
   height: 100%;
 `;
+
+export const ErrorText = styled.p`
+  color: ${COLORS.redMid};
+  font-size: 0.875rem; //14px
+  font-weight: 400;
+  ${openSans.style}
+`;
```
jinkang-0: can we use the `P` text style component with `$color={COLORS.redMid}` instead?

src/api/supabase/queries/interest.ts
```diff
@@ -12,8 +12,8 @@ export async function getAllInterests() {
 }
 
 /** Insert a new interest object into supabase interest table. */
```
jinkang-0: we might want to update this comment lol

jinkang-0 (commented):
looks great!! i dropped some nits and notes on the code, ill start reviewing the website soon

src/components/DateInput/styles.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+  return COLORS.greyLight;
+};
+
+export const InputLabel = styled(H4)`
+  color: ${COLORS.greyDark};
+  margin-bottom: 0.625rem; //10px
+`;
+
+export const InputDate = styled.input<{ $error: boolean; $filled: boolean }>`
```
jinkang-0: i think we should add a `cursor: text` to the date input to stay consistent with other text-input-like designs

src/components/InterestForm/index.tsx
```diff
@@ -28,89 +30,110 @@ export default function InterestForm({ caseData }: { caseData: CaseListing }) {
...
+              <Radio key={option}>
+                <RadioInput
+                  id={`radio${option}`}
+                  type="radio"
+                  name="radioOptions"
+                  value={option}
+                  checked={rolesInterested === option}
+                  onChange={event => setRolesInterested(event.target.value)}
+                />
+                <RadioLabel>{option}</RadioLabel>
```
jinkang-0: I'd recommend adding a `htmlFor` prop to the labels that match the `id` of the `RadioInput` they correspond to, which will let you select the radio when you click on the label.

src/components/Button.tsx
```diff
@@ -4,41 +4,42 @@ import styled from 'styled-components';
...
+      $primaryColor (Omit) - IMPORTANT: DO NOT PASS IN A primaryColor PROP!!! The background color will be white by default
+      $secondaryColor (Required) - This color will determine the border color and background color on hover
 
     Example: 
-      <Button secondaryColor={COLORS.primaryBlue} onClick={...}>
+      <Button $secondaryColor={COLORS.blueMid} onClick={...}>
         [Button text here]
       </Button>
  */
 const Button = styled.button<{
```
jinkang-0: we may want a `cursor: pointer` to the button

src/components/Button.tsx
```diff
@@ -4,41 +4,42 @@ import styled from 'styled-components';
...
-  secondaryColor: string;
+  $primaryColor?: string;
+  $secondaryColor: string;
 }>`
   appearance: none;
-  color: ${props => (props.primaryColor ? 'white' : 'black')};
-  background: ${props => (props.primaryColor ? props.primaryColor : 'white')};
+  color: ${props => (props.$primaryColor ? 'white' : 'black')};
+  background: ${props => (props.$primaryColor ? props.$primaryColor : 'white')};
   align-self: flex-end;
```
jinkang-0: since the regular button will be used throughout the app, should we remove the `align-self` property? it may be best to let the container of the button control its positioning

jinkang-0:
it looks like i was able to submit an interest with a date before the current day by typing the values out

image

we may want to add some value validation before submitting lol

jinkang-0:
this is out of scope, but tab and shift + tab navigation seems to be broken on the radio group. not too sure why, but could experiment for a fix in the future

src/components/Button.tsx
jinkang-0: also `outline: none` to remove the outline when focused

src/components/Button.tsx
```diff
@@ -4,41 +4,42 @@ import styled from 'styled-components';
...
   padding: 0.625rem 1.25rem;
   border-radius: 0.313rem; // 5px
   border: 2px solid
-    ${props => (props.primaryColor ? props.primaryColor : props.secondaryColor)};
+    ${props =>
+      props.$primaryColor ? props.$primaryColor : props.$secondaryColor};
   &:hover {
```
jinkang-0: `&:hover` styles should also be applied to the `&:focus` state

src/components/InterestForm/styles.tsx
```diff
@@ -28,34 +28,32 @@ export const FormTextArea = styled.textarea`
...
+  &:not(:placeholder-shown) {
+    border-color: ${COLORS.greyMid};
   }
 `;
 
 export const RadioGroup = styled.div`
   display: flex;
   flex-direction: row;
-  margin-bottom: 0.6rem;
+  align-self: center;
```
pragyakallanagoudar: think this might be causing the set of radio buttons to be shifted to the right horizontally? 

Screenshot 2023-11-23 at 3 35 38 PM

src/components/DateInput/index.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { Dispatch, SetStateAction } from 'react';
+import { InputLabel, ErrorText, InputDiv, InputDate } from './styles';
+
+type TextInputProps = {
```
pragyakallanagoudar: nit: can we change this to `DateInputProps`

src/components/InterestForm/index.tsx
```diff
@@ -28,89 +30,110 @@ export default function InterestForm({ caseData }: { caseData: CaseListing }) {
...
+            ))}
+            {missingInfo && rolesInterested === '' && (
+              <ErrorText>Must select your preferred role</ErrorText>
+            )}
+          </RadioGroup>
+          <DateInput
+            label="What is the earliest date you can contact the client?"
+            error={missingInfo && startDate === ''}
+            filled={startDate !== ''}
+            errorText="Must include earliest date"
```
pragyakallanagoudar: nit: can we make this "Must include earliest contact date"

src/components/InterestForm/index.tsx
```diff
@@ -28,89 +30,110 @@ export default function InterestForm({ caseData }: { caseData: CaseListing }) {
...
+      setMissingInfo(true);
+      return;
+    }
+    if (startDate !== '' && rolesInterested !== '') {
       const newInterest: Interest = {
         // hardcoded values for now
         listing_id: caseData.id,
         listing_type: 'CASE',
-        user_id: '36b8f84d-df4e-4d49-b662-bcde71a8764f',
+        user_id: '515d9b33-9185-4601-9a13-23789bea3ac0',
```
pragyakallanagoudar, line 56: next step here is that we want to pull the authenticated user's id from supabase, but let's handle that in the final interest form sprint (!!!!!!!!) which will have that, un-authenticated/un-onboarded states, and the profile match stuff

pragyakallanagoudar (commented):
LOOKS SO GOOD ALVAROOOO!! i made a bunch of form submissions and the supabase seems to reflect them perfectly. 

i think there's just a couple of small nits / suggestions to address, but then should be good to go :DD

src/components/DateInput/index.tsx
varortz: oh ye, I think I just forgot to delete this

src/components/DateInput/index.tsx
varortz: lol yes ur right, idk why I just use `value` to begin with

src/components/DateInput/styles.tsx
varortz: Note [worksesh]: moving forward we should go with the destructuring approach

src/components/Button.tsx
varortz: Note [worksesh]: moving forward with keeping the outline for the `Button` component since adding styles to `&:focus` state results in weird behavior. Specifically, after clicking the button, it will remain in focus state. When the form has errors, it kinda looks like the button is stuck. Discussed with @davidqing6432 that we should not add a custom `outline-color` and allow for the outline color to default to whatever the user's outline color is on their device.

src/components/DateInput/styles.tsx
varortz: Note: done by jinkang, ty!

src/utils/helpers.ts
```diff
@@ -36,9 +36,12 @@ export const parseAgency = (agency: string): string =>
  * @returns true if the date is an upcoming date, false otherwise
  */
 export const isValidDate = (d: string) => {
-  const currentDate = new Date();
-  currentDate.setHours(0, 0, 0, 0);
-  if (d !== '' && new Date(d) >= currentDate) {
+  const tomorrow = new Date();
```
pragyakallanagoudar: question: do we not want them to be able to select the current day? i feel like if the user has no restrictions on what day they can contact the client, it would feel more intuitive to select the current day as opposed to the next day

varortz: oh yeah lemme update, I agree. I should probably update the wording for the error feedback too! Edit: Should be updated now

PR #51, opened 2023-12-02
[feat] create legal experience card

ccatherinetan:

`src/app/onboarding/legal-experience`
Creates the legal experience card, updating the profile with Bar Number, Legal Experience, and EOIR registration, if all inputs are valid.
The Continue button is enabled at the same time as updateProfile is called. 

Valid Inputs → Enabled Continue
Screen Shot 2023-12-02 at 2 37 24 PM

Invalid Bar Number → Disabled Continue 
Screen Shot 2023-12-02 at 2 37 17 PM
Screen Shot 2023-12-02 at 2 34 36 PM

nav to http://localhost:3000/onboarding/legal-experience
* check Disabled button if 
    * non-numeric bar number input 
    * bar number input length != 6
    * EOIR registration or Legal Experience not selected 

## :seedling: Next steps 

CC: @pragyakallanagoudar

### DIFF
--- a/src/app/onboarding/legal-experience/page.tsx
+++ b/src/app/onboarding/legal-experience/page.tsx
@@ -38,8 +38,16 @@ export default function Page() {
     if (isRegistered !== undefined)
       setRegistered(onboarding.profile.eoir_registered ? 'Yes' : 'No');
 
-    if (isValidBarNumber(barNumber) && experience && isRegistered !== undefined)
+    if (
+      isValidBarNumber(barNumber) &&
+      experience &&
+      isRegistered !== undefined
+    ) {
       onboarding.setCanContinue(true);
+    } else {
+      onboarding.setProgress(3);
+      onboarding.setCanContinue(false);
+    }
   }, [onboarding, isValidBarNumber]);
 
   const handleBarNumChange = useCallback(
--- a/src/app/onboarding/legal-experience/page.tsx
+++ b/src/app/onboarding/legal-experience/page.tsx
@@ -14,12 +14,6 @@ const legalExperienceOptions = new Map<ImmigrationLawExperienceEnum, string>([
   ['LOW', 'One or no case of immigration law experience'],
 ]);
 
-// const ExperienceTypes = new Set([
-//   'Multiple cases of immigration law experience',
-//   'Few cases of immigration law experience',
-//   'One or no case of immigration law experience',
-// ]);
-
 export default function Page() {
   const onboarding = useContext(OnboardingContext);
   const [barNum, setBarNum] = useState('');
@@ -38,7 +32,7 @@ export default function Page() {
   };
 
   useEffect(() => {
-    if (barNum !== '' && experience !== null && registered !== '') {
+    if (validBarNum() && experience !== null && registered !== '') {
       // update profile
       const partialProfile: Partial<Profile> = {
         bar_number: barNum,
@@ -47,9 +41,7 @@ export default function Page() {
       };
       onboarding?.updateProfile(partialProfile);
       // enable continue
-      if (validBarNum()) {
-        onboarding?.setCanContinue(true);
-      }
+      onboarding?.setCanContinue(true);
     } else {
       onboarding?.setCanContinue(false);
     }
--- a/src/app/onboarding/legal-experience/page.tsx
+++ b/src/app/onboarding/legal-experience/page.tsx
@@ -1,22 +1,25 @@
 'use client';
 
 import { useContext, useState, useEffect } from 'react';
-import Button from '@/components/Button';
-import COLORS from '@/styles/colors';
 import { H1 } from '@/styles/text';
 import { OnboardingContext } from '@/utils/OnboardingProvider';
 import TextInput from '@/components/TextInput';
 import InputDropdown from '@/components/InputDropdown';
+import RadioGroup from '@/components/RadioGroup';
 import { ImmigrationLawExperienceEnum, Profile } from '@/types/schema';
 
-const ExperienceTypes = new Set(['LOW', 'MEDIUM', 'HIGH']);
+const ExperienceTypes = new Set([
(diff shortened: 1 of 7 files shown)


src/app/onboarding/legal-experience/page.tsx
```diff
@@ -1,24 +1,78 @@
...
 
 export default function Page() {
   const onboarding = useContext(OnboardingContext);
+  const [barNum, setBarNum] = useState('');
+  const [experience, setExperience] =
+    useState<ImmigrationLawExperienceEnum | null>(null);
+  const [registered, setRegistered] = useState('');
+
+  const validBarNum = () =>
+    !Number.isNaN(parseInt(barNum, 10)) && barNum.length === 6;
```
jinkang-0: we can probably turn this into a variable with `useMemo` instead of a function

src/app/onboarding/legal-experience/page.tsx
```diff
@@ -1,24 +1,78 @@
...
+
+  const getErrorText = () => {
+    if (barNum !== '' && !validBarNum()) {
+      return 'Must include attorney bar number';
+    }
+    return '';
+  };
```
jinkang-0: can we turn this into a `useMemo` as well?

src/app/onboarding/legal-experience/page.tsx
jinkang-0: also bar number can be less than 6 digits

src/app/onboarding/legal-experience/page.tsx
```diff
@@ -1,24 +1,78 @@
...
+        immigration_law_experience: experience,
+        eoir_registered: registered === 'Yes',
+      };
+      onboarding?.updateProfile(partialProfile);
+      // enable continue
+      onboarding?.setCanContinue(true);
+    } else {
+      onboarding?.setCanContinue(false);
+    }
+  }, [barNum, experience, registered]);
```
jinkang-0: instead of a `useEffect`, I would recommend making individual `onChange` functions for each input (I've added a separate `onChange` prop to each of them that triggers a side effect)

jinkang-0: This is so we can populate the data by grabbing values from `onboarding` with `useEffect` -- if we keep the one that we have here, it will result in an infinite recursive loop (since they update values in each others' dependency arrays)

src/app/onboarding/legal-experience/page.tsx
```diff
@@ -1,24 +1,97 @@
...
 
 export default function Page() {
   const onboarding = useContext(OnboardingContext);
+  const [barNum, setBarNum] = useState('');
+  const [registered, setRegistered] = useState('');
+  const [barNumError, setBarNumError] = useState('');
+
+  const isValidBarNumber = useCallback(
+    (barNumber: string) =>
+      !Number.isNaN(parseInt(barNumber, 10)) && /^\d{6}$/.test(barNumber),
```
jinkang-0: for the regex, we may want to consider `/^\d{1,6}$/` instead, since a bar number can be less than 6 digits

src/app/onboarding/legal-experience/page.tsx
```diff
@@ -1,24 +1,97 @@
...
+      setRegistered(onboarding.profile.eoir_registered ? 'Yes' : 'No');
+
+    if (
+      isValidBarNumber(barNumber) &&
+      experience &&
+      isRegistered !== undefined
+    ) {
+      onboarding.setCanContinue(true);
+    } else {
+      onboarding.setProgress(3);
```
jinkang-0, line 42: ideally we shouldn't manually set progress, it is already handled by the onboarding manager

src/app/onboarding/legal-experience/page.tsx
```diff
@@ -1,24 +1,97 @@
...
+      onboarding.setCanContinue(false);
+    }
```
jinkang-0, line 44: can we can turn this into a ternary statement?

pragyakallanagoudar: hmm how do we make it ternary given that the else clause has 2 lines?

src/app/onboarding/legal-experience/page.tsx
pragyakallanagoudar: another source i found said 6-8 digits? i'm going to do that, but i think none of these are comprehensive enough (ChatGPT tells me that some bar numbers in some states can have letters) so i think ultimately we want something more robust.

for some reason the info online is really sparse though, so idt getting this perfectly right is feasible time-wise rn (added to [cleanup](https://www.notion.so/calblueprint/CLEANUP-381a4634fcd84ce2abad62dd9c8a6470))

src/app/onboarding/legal-experience/page.tsx
jinkang-0: Disregard this lol, was considering removing progress
could look back when we figure out a better solution for onboarding progress

src/app/onboarding/legal-experience/page.tsx
jinkang-0: Here is what I can see from the California state bar website: https://apps.calbar.ca.gov/attorney/LicenseeSearch/QuickSearch?FreeText=Smith

It looks like it can go as low as 3 digits to 6 digits

jinkang-0: Definitely can be improved, but we need to check with the NPO on that. I think for California (our scope) though, it shouldn't be 6 or more.

pragyakallanagoudar: ooh ok agreed! im just going to change to 1-6 characters and we can approach this again later. resolving for rn, thanks for doing this research jinkang!

jinkang-0 (approved):

pragyakallanagoudar (approved):

PR #54, opened 2023-12-03
[feat] creates basic information card

pragyakallanagoudar:

This PR creates the basic information card in the onboarding flow. It includes `TextInput` components for the first and last name, and `BigDataDropdown` components for the city, languages spoken, and languages read. 

On editing any field, the corresponding field in the `Partial<Profile>` object in the `OnboardingContext` instance is updated. The "Continue" button is not enabled until all fields related to this card have been filled in. 

Screenshot 2023-12-05 at 10 26 34 AM

Screenshot 2023-12-05 at 10 27 11 AM

1. Enter the onboarding flow and select a role.
2. You should now be able to view the `/onboarding/basic-information` route.
3. Fill out some of the information, and make sure that the "Continue" button is not enabled.
4. Now fill out all fields, and press "Continue".
5. Use the back button to route back to `/onboarding/basic-information`, and make sure the fields are prefilled with what you originally filled out. 

## :seedling: Next steps 

CC: @pragyakallanagoudar


src/app/onboarding/basic-information/page.tsx
```diff
@@ -1,24 +1,92 @@
...
+
+  useEffect(() => {
+    if (
+      firstName !== '' &&
+      lastName !== '' &&
+      onboarding?.profile.location !== undefined
+    ) {
+      onboarding?.setCanContinue(true);
+    } else {
+      onboarding?.setProgress(1);
```
jinkang-0, line 28: ideally we shouldn't set progress manually, it is already handled by the onboarding manager

src/app/onboarding/basic-information/page.tsx
```diff
@@ -1,24 +1,92 @@
...
 import { OnboardingContext } from '@/utils/OnboardingProvider';
-import { useContext } from 'react';
+import TextInput from '@/components/TextInput';
+import { LineDiv } from './styles';
 
 export default function Page() {
   const onboarding = useContext(OnboardingContext);
+  const [firstName, setFirstName] = useState('');
+  const [lastName, setLastName] = useState('');
+  useEffect(() => {
```
jinkang-0, line 16: can we add a line before the `useEffect` lol

src/app/onboarding/basic-information/page.tsx
```diff
@@ -1,24 +1,92 @@
...
+    setFirstName(onboarding?.profile.first_name || '');
+    setLastName(onboarding?.profile.last_name || '');
+  }, []);
```
jinkang-0: since onboarding is used in the `useEffect`, it should be listed as a dependency in the dependencies array

src/app/onboarding/basic-information/page.tsx
```diff
@@ -1,24 +1,92 @@
...
+      firstName !== '' &&
+      lastName !== '' &&
+      onboarding?.profile.location !== undefined
+    ) {
+      onboarding?.setCanContinue(true);
+    } else {
+      onboarding?.setProgress(1);
+      onboarding?.setCanContinue(false);
+    }
+  }, [onboarding, firstName, lastName]);
```
jinkang-0: Instead of setting `canContinue` here using state values, I would recommend checking it against what we have in the onboarding context -- this ensures that what will eventually be pushed to supabase is accurate (and also does not cause a bunch of recursive `useEffects`)

src/app/onboarding/basic-information/page.tsx
```diff
@@ -1,24 +1,92 @@
...
+
+  useEffect(() => {
+    onboarding?.updateProfile({
+      first_name: firstName,
+      last_name: lastName,
+    });
+  }, [firstName, lastName]);
```
jinkang-0: I would recommend using `onChange` event listeners instead, like what we are doing already for City and languages

src/app/onboarding/basic-information/page.tsx
jinkang-0: also we should check for location and languages for `canContinue`. Note that location is defaulted to `''` in the onboarding context, so it is never undefined. I would recommend using its truthy value instead.

src/app/onboarding/basic-information/page.tsx
pragyakallanagoudar: ahh the issue here is that if you move forward to a different card, come back to basic info, and then empty out a required field, the continue button will still be enabled

jinkang-0: oh i see, then it's fine

pragyakallanagoudar: which is why i'm manually forcing the progress back

...also curious, why do we need the progress? can't we just always set the `canContinue` attribute based on whether the fields are populated, which we can pull from the onboarding context 🤔

jinkang-0: the progress is there mainly as a security measure so people can't just enter in the url and skip steps

jinkang-0: and also keeps track of well.. the progress, on the progress bar.

jinkang-0: prolly a better way to do this though

src/app/onboarding/basic-information/page.tsx
pragyakallanagoudar: ahhh i think this causes a circular dependency as @ccatherinetan identified?

jinkang-0: It won't if we switch to using onChange and remove the other `useEffect`

src/app/onboarding/basic-information/page.tsx
pragyakallanagoudar: for some reason when i tried this it did not work, but im going to play around with it again!!

src/app/onboarding/basic-information/page.tsx
pragyakallanagoudar: updated to check the location correctly, but personally im not sure if languages should be required? cc @davidqing6432, also jinkang curious to hear your thoughts on that

src/app/onboarding/basic-information/page.tsx
pragyakallanagoudar: nvm this hella works am pushing

src/app/onboarding/basic-information/page.tsx
pragyakallanagoudar: actually...i change my mind LOL let's make these required bc ambreen said we can assume all volunteers can read/write in either english or spanish

src/app/onboarding/basic-information/page.tsx
pragyakallanagoudar: so true, pushed

src/app/onboarding/basic-information/page.tsx
```diff
@@ -12,30 +12,35 @@ export default function Page() {
...
 
-  useEffect(() => {
+  const handleFirstName = (v: string) => {
+    onboarding?.updateProfile({
+      first_name: v,
+    });
+  };
+
+  const handleLastName = (v: string) => {
     onboarding?.updateProfile({
```
jinkang-0: can we call these `handleFirstNameChange` and `handleLastNameChange` respectively?

jinkang-0: also can we just define them in the props to the input elements?

pragyakallanagoudar: added "Change"! can i veto the prop idea lol this just feels clearer

jinkang-0: i feel like we should stay consistent though, if we're putting this here, we should also make a separate `onChange` function for the dropdown

jinkang-0 (approved):

PR #60, opened 2023-12-05
[fix] address not logged in and already onboarded states in onboarding flow

jinkang-0:

- Modified `components/OnboardingManager.tsx` to redirect if user is not logged in or has already completed onboarding.
- Added `lib/config.ts` to make future edit to home page URL easier

## :seedling: Next steps 

CC: @varortz

### DIFF
--- a/src/components/OnboardingManager.tsx
+++ b/src/components/OnboardingManager.tsx
@@ -34,6 +34,7 @@ export default function OnboardingManager({
     if (
       onboarding.progress < 0 ||
       onboarding.progress >= onboarding.flow.length ||
+      !profile?.userId ||
       (profile?.profileReady && profile?.profileData)
     ) {
       router.push(CONFIG.homepage);
--- a/src/components/OnboardingManager.tsx
+++ b/src/components/OnboardingManager.tsx
@@ -27,26 +27,22 @@ export default function OnboardingManager({
   }, [pathname, onboarding]);
 
   useEffect(() => {
-    if (!profile) return;
-
-    if (!profile.userId || (profile.profileReady && profile.profileData))
-      router.push('/cases');
-  }, [profile, router]);
-
-  useEffect(() => {
     if (!onboarding) return;
 
     // out of bounds redirect
     if (
       onboarding.progress < 0 ||
-      onboarding.progress >= onboarding.flow.length
+      onboarding.progress >= onboarding.flow.length ||
+      profile?.userId ||
+      (profile?.profileReady && profile?.profileData)
     ) {
       router.push('/cases');
+      return;
     }
 
     if (pageProgress > onboarding.progress)
       router.push(`/onboarding/${onboarding.flow[onboarding.progress].url}`);
-  }, [onboarding, router, pageProgress]);
+  }, [onboarding, profile, router, pageProgress]);
 
   const advanceProgress = () => {
     // safeguard
--- a/src/components/OnboardingManager.tsx
+++ b/src/components/OnboardingManager.tsx
@@ -6,6 +6,7 @@ import { OnboardingContext } from '@/utils/OnboardingProvider';
 import { usePathname, useRouter } from 'next/navigation';
 import { ReactNode, useContext, useEffect, useMemo } from 'react';
 import { ProfileContext } from '@/utils/ProfileProvider';
+import CONFIG from '@/lib/configs';
 import ProgressBar from './ProgressBar';
 import BigButton from './BigButton';
 
@@ -33,10 +34,9 @@ export default function OnboardingManager({
     if (
       onboarding.progress < 0 ||
       onboarding.progress >= onboarding.flow.length ||
-      profile?.userId ||
       (profile?.profileReady && profile?.profileData)
     ) {
-      router.push('/cases');
+      router.push(CONFIG.homepage);
       return;
     }
 
@@ -61,7 +61,7 @@ export default function OnboardingManager({
 
     if (newProgress >= onboarding.flow.length) {
       onboarding.flushData().then(() => {
-        router.push('/cases');
+        router.push(CONFIG.homepage);
       });
     } else {
       router.push(`/onboarding/${onboarding.flow[newProgress].url}`);
@@ -74,7 +74,7 @@ export default function OnboardingManager({
         href={
           onboarding && pageProgress > 0
             ? `/onboarding/${onboarding.flow[pageProgress - 1].url}`
-            : '/cases'
+            : CONFIG.homepage
         }
       >
         Back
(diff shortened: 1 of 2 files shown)


src/lib/configs.ts
```diff
@@ -0,0 +1,5 @@
+const CONFIG = {
+  homepage: '/cases',
```
varortz: would it be possible to update the routing to the homepage so we don't have to update this later?

varortz (approved):
Looks good to merge! THANK U JINKANG

src/components/OnboardingManager.tsx
```diff
@@ -30,14 +33,17 @@ export default function OnboardingManager({
     // out of bounds redirect
     if (
       onboarding.progress < 0 ||
```
ccatherinetan, line 35: when would  `onboarding.progress` be less than 0? 
edit: is this related to line 22-23? 
` const pageProgress = useMemo(() => {
    if (!onboarding) return -1;`

jinkang-0: ~~when the user presses the back button on the role page~~
see below

src/components/OnboardingManager.tsx
```diff
@@ -30,14 +33,17 @@ export default function OnboardingManager({
...
-      onboarding.progress >= onboarding.flow.length
+      onboarding.progress >= onboarding.flow.length ||
+      !profile?.userId ||
+      (profile?.profileReady && profile?.profileData)
```
ccatherinetan, line 38: everything looks good! i just have some clarification questions 
when is `profile.profileReady` true? If it's only true when the entire profile is filled out, do we need to check `profileData` too? 
i'm assuming this line is to reroute when the user has already filled out onboarding?

src/components/OnboardingManager.tsx
```diff
@@ -30,14 +33,17 @@ export default function OnboardingManager({
...
     ) {
-      router.push('/cases');
+      router.push(CONFIG.homepage);
+      return;
     }
 
     if (pageProgress > onboarding.progress)
       router.push(`/onboarding/${onboarding.flow[onboarding.progress].url}`);
-  }, [onboarding, router, pageProgress]);
+  }, [onboarding, profile, router, pageProgress]);
```
ccatherinetan, line 46: just to clarify, why is it necessary to have profile in the dependency list?
it is b/c the profile will change as the user goes through onboarding?

ccatherinetan (commented):
i have some remaining questions, but it's almost good to go!

src/components/OnboardingManager.tsx
jinkang-0: `profile.profileReady` is set to true after we fetch profile data from Supabase.
`profile.profileData` is what we get from Supabase.

Both are included to check:
1) we have the profile data from the user
2) the user has a profile (i.e. has already been onboarded)

If both are true, then we want to redirect the user because they have already finished onboarding (i.e. the goal of this PR).

src/components/OnboardingManager.tsx
jinkang-0: In general, it is good practice to include all the variables that are included in the function of `useEffect` in the dependency array so the function is ran again when the variable updates.

Practically, one example scenario is that the user can load the website on any onboarding page. Since the app begins on a cold start, we would need to wait for the profile data to be fetched. In the meantime, the page would've already loaded, so we would have to re-run this function once we have the profile data fetched. To tell React to do this, we add `profile` to the dependency array.

src/components/OnboardingManager.tsx
jinkang-0: to the edit: that is different.

`pageProgress` records the progress corresponding to the current page. `onboarding.progress` records the overall progress. For example, they may have filled out up to `Legal Experience`, but want to go back to edit `Basic Information`. In this case, `pageProgress=1`, `onboarding.progress=3`.

The `-1` is simply an indication that something has gone wrong, since the `OnboardingManager` must have access to the `OnboardingContext`. In hindsight, there may be a better way to indicate this (i.e. through tossing an error), but that would require more refactoring that's irrelevant to the goal of this PR. It's also not possible to happen since we guarantee `OnboardingManager` to be defined inside an `OnboardingContext`. The error check is more for when we want to export this component to another usage (which we won't do).

Update: it looks like the `profile.progress < 0` check is more just a safe guard for something that would likely never happen. In practice, I don't think we'll ever decrease `profile.progress`, so this is probably never going to happen. That being said, it won't hurt to leave it in lol.

ccatherinetan: sg, thanks for the clarification. i'm not sure what the code cleanliness/readibility standards are, but if you think it's fine to leave it in, i'm fine with it too. i guess the argument for taking it out would be that it could prevent confusion on the matter if someone else sees it?

src/components/OnboardingManager.tsx
ccatherinetan: sounds good, i guess i'm just wondering if it's enough to check that we have profile data, i.e. just `profile?.profileData`. Doesn't this imply that the profile is ready?

src/components/OnboardingManager.tsx
ccatherinetan: ahhh ic, that makes sense; thank you!

src/components/OnboardingManager.tsx
jinkang-0: No, because `profileData` will be undefined before being populated and is therefore falsy. Perhaps we can make an argument to check `undefined` vs `null` to discern between "not initialized" vs "initialized but empty", but I don't think we have made a clear decision on whether to use that difference or not. More importantly, the ability to use `null` and `undefined` depends on the underlying database. Even if Supabase gives us `null` for an empty record, other databases may choose to represent empty items with `undefined`, and this could lead to a mysterious bug if the tech team decides to switch databases after hand off. `profileReady` is an explicit boolean that we know we have set and can check to make sure the data is ready.

src/components/OnboardingManager.tsx
jinkang-0: If the source of confusion is the `onboarding.progress < 0` check, I'd say just treat it as a sanity check. It shouldn't be less than 0, and if it is, something went wrong, so redirect. It's harmless as is, and the utility is marginal, but for the sake of my sanity, I believe it's better to have it than not.

jinkang-0: But can you elaborate why you think this is a readability issue or source of confusion? Intuitively, I naturally think of an error boundary / sanity check when I see a `< 0` check for something that realistically shouldn't be negative and won't pay it too much mind, but I'm not sure if I see it the same way as you.

ccatherinetan (approved):
looks good!

PR #63, opened 2024-03-02
[feat] Setup Vercel function to update database

jinkang-0:

- Added an API route to update the database (only works with correct keys)
- Added CRON job to run the function 5:00 AM UTC every day (only works when deployed)

In `.env.local`, define:
```
LEGAL_SERVER_URL=<the API endpoint of legal server>
LEGAL_SERVER_AUTHORIZATION=<BP bearer ID for legal server>
SUPABASE_SERVICE_KEY=<Supabase service key to update database>
```

Then navigate to `/api/legal-db` to run the function. Check Supabase for updates.
If it looks like nothing changed, try deleting something and run it again.

## :seedling: Next steps 

Setup rate limiting.

CC: @varortz

### DIFF
--- a/src/app/api/legal-db/route.ts
+++ b/src/app/api/legal-db/route.ts
@@ -1,0 +1,117 @@
+import { CaseListing } from '@/types/schema';
+import { createClient } from '@supabase/supabase-js';
+
+export const dynamic = 'force-dynamic';
+
+type CasesData = Omit<Omit<CaseListing, 'relief_codes'>, 'languages'>;
+
+interface CaseLanguage {
+  listing_id: string;
+  language_name: string;
+}
+
+interface CaseRelief {
+  listing_id: string;
+  relief_code: string;
+}
+
+interface ReportData {
+  id: string;
+  matter_builtin_lookup_case_disposition_case_disposition_expn: string;
+  matter_builtin_lookup_case_status_case_status_expn: string;
+  program_name: string;
+  matter_builtin_lookup_problem_code_legal_problem_code_expn: string;
+  identification_number: string;
+  person_builtin_lookup_language_language_expn: string;
+  city: string;
+  matter_builtin_lookup_country_country_of_origin_expn: string;
+  state: string;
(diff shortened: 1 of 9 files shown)


src/app/api/legal-db/route.ts
```diff
@@ -0,0 +1,168 @@
+import { CaseListing } from '@/types/schema';
```
pragyakallanagoudar, line 1: why are we creating an additional `api` folder in `app`? feels a bit confusing to have `src/api` and `src/app/api`

pragyakallanagoudar: NEVERMIND this is for routing...

src/components/ListingCard/index.tsx
```diff
@@ -18,7 +17,7 @@ export default function ListingCard({
 }: {
   caseData: CaseListing;
   isSelected?: boolean;
-  onClick?: (id: UUID) => void;
+  onClick?: (id: string) => void;
 }) {
   // Helper function to generate an array of strings for CardTag components
   const generateCardTags = (): string[] => {
```
varortz, line 23: looks like the bug/console error in the screenshot below was introduced because `caseData.relief_codes` and `caseData.languages` are both lists with one null entry. This results in empty tags on the `ListingCard` component, even when there are no languages or relief codes associated with the case listing. Could we try getting rid of these console log errors? If you want a quick solution we can update the `if` checks on lines 34 and 42 to check that the 0th of lists does not equal null (`caseData.languages[0] !== null`), but there might be a less hacky fix? image

jinkang-0: i believe that's out of scope for this sprint, but i can include that for now
it would be better if we do a full fix of the cases page as another sprint.

varortz: THANK U FOR THE FIX!! But yeah, I agree that we'll need to make some fixes to the cases page in future sprints 👍 just wanted to avoid introducing console errors into the main branch

jinkang-0:
switching to retool workflow

PR #37, opened 2023-11-12
[chore] modify profile context

jinkang-0:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

### :palm_tree: What's new in this PR
[//]: # "REQUIRED - Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."

Modified `src/utils/ProfileProvider.tsx` to track profile languages and roles.
- Added context props `languages`, `roles` to provide their values
- Added context props `setLanguages`, `setRoles` to allow updating values (overwrites them entirely)
    - Note: requires RLS policies

Modified `src/api/supabase/queries/profiles.ts` to support relevant CRUD operations.

Unplanned edits:
- Modified `.gitignore` to ignore `src/app/test/` for convenience in testing

### :evergreen_tree: Screenshots
[//]: # "REQUIRED for frontend changes, otherwise optional but strongly recommended. Add screenshots of expected behavior - GIFs if you're feeling fancy! If you are making changes here, please CC: @davidqing6432 at the bottom."

## :deciduous_tree: How to review
[//]: # 'REQUIRED - Describe the order in which to review files and what to expect when testing locally. Is there anything specifically you want feedback on? Should this be reviewed commit by commit, or all at once? What are some user flows to test? What are some edge cases to look out for?'

#### Code Review

First review changes in `src/api/supabase/queries/profiles.ts`, then `src/utils/ProfileProvider.tsx`.

#### Code Testing

Standard procedure:
```
git fetch jinkang/ijp-66-modify-profile-context
git checkout jinkang/ijp-66-modify-profile-context
```

Pre-reqs before testing:
- Follow usage in `src/utils/ProfileProvider.tsx` to setup testing page
- Sign in or log in with a test account
- Ensure you have a corresponding row with the correct `user_id` field in the `profiles` table

If all of the above are done, navigate to the test page and mess around.

## :seedling: Next steps 
[//]: # "Optional - What's NOT in this PR, doesn't work yet, and/or still needs to be done. Note any temporary fixes in this PR that should be cleaned up later."

### :information_source: Online sources
[//]: # 'Optional - copy links to any tutorials or documentation that was useful to you when working on this PR'

### :potted_plant: Related PRs
[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc; if this is a refactor, feel free to add PRs that previously modified this code"

[//]: # 'This tags the project leader as a default. Feel free to change, or add on anyone who you should be in on the conversation.'
[...]

PR #23, opened 2023-10-22
[feat] add profile context and provider

jinkang-0:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

### :palm_tree: What's new in this PR
[//]: # "REQUIRED - Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."

Added `src/utils/ProfileProvider.tsx` to create the context and context provider.
- Check comments at the bottom of the file for usage

Modified `src/api/supabase/queries/profiles.ts` to include a helper function to fetch profile by id.

### :evergreen_tree: Screenshots
[//]: # "REQUIRED for frontend changes, otherwise optional but strongly recommended. Add screenshots of expected behavior - GIFs if you're feeling fancy! If you are making changes here, please CC: @davidqing6432 at the bottom."

## :deciduous_tree: How to review
[//]: # 'REQUIRED - Describe the order in which to review files and what to expect when testing locally. Is there anything specifically you want feedback on? Should this be reviewed commit by commit, or all at once? What are some user flows to test? What are some edge cases to look out for?'

Review in any order.

## :seedling: Next steps 
[//]: # "Optional - What's NOT in this PR, doesn't work yet, and/or still needs to be done. Note any temporary fixes in this PR that should be cleaned up later."

### :information_source: Online sources
[//]: # 'Optional - copy links to any tutorials or documentation that was useful to you when working on this PR'

### :potted_plant: Related PRs
[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc; if this is a refactor, feel free to add PRs that previously modified this code"

[//]: # 'This tags the project leader as a default. Feel free to change, or add on anyone who you should be in on the conversation.'
CC: @pragyakallanagoudar

PR #64, opened 2024-03-03
[feat] creates the auth context

pragyakallanagoudar:

- Adds the `AuthContext`, which has the (1) user session, (2) user ID, (3) user email, (4) sign in function, (5) sign out function, and (6) sign up function for use across the application.
- Removes the user ID and user email from `ProfileContext`, and instead modifies `ProfileContext` to retrieve user ID from the `AuthContext`.
- Refactors `/login`, `/signup`, and the sign out button functionality to use the `AuthContext` as opposed to calling the Supabase functions directly. 
- Refactors the components that originally used the user ID from the `ProfileContext` to use the `AuthContext` for this instead.
- **NOTE:** This does not refactor the `/reset-password` code to use the `AuthContext`, will discuss more tomorrow but I think we should probably look at the issues with auth emails first since we can't test reset password until that's resolved.

N/A, all pages are the same visually

- Look at the newly created `src/utils/AuthProvider` and the changes to `src/utils/ProfileProvider`.
- Look at the changes to the auth pages/functionality on `/login`, `/signup`, and `/cases`. 
- Look at the changes in `src/components`.
- Run the app and test all auth-related functionality (from the previous two bullet points).
   - Before you do this, comment out anything in the `CaseDetails` component that uses legal experience -- this is no longer in the schema and so it errors.
- ❗**THINGS I'M CONFUSED ABOUT (help)**: 
   - Any modified dependencies for `useCallback`, `useMemo`, and `useEffect`
   - Anything marked with a TODO

## :seedling: Next steps 

- Refactor the `/reset-password` code to use the `AuthContext`.

## :link: Relevant Links

https://github.com/calblueprint/girls-write-now/blob/main/src/utils/AuthContext.tsx (huge adi ++)

N/A

#23 #37 (huge jinkang ++)

CC: @varortz

### DIFF
--- a/src/app/(auth)/signup/page.tsx
+++ b/src/app/(auth)/signup/page.tsx
@@ -23,6 +23,7 @@ export default function SignUp() {
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
   const handleSignUp = async () => {
+    console.log(auth);
     if (!auth) return;
     setEmailError(validEmail(email) ? '' : 'Invalid Email');
     setPasswordError(password !== '' ? '' : 'Invalid Password');
--- a/src/app/(auth)/signup/page.tsx
+++ b/src/app/(auth)/signup/page.tsx
@@ -23,7 +23,6 @@ export default function SignUp() {
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
   const handleSignUp = async () => {
-    console.log(auth);
     if (!auth) return;
     setEmailError(validEmail(email) ? '' : 'Invalid Email');
     setPasswordError(password !== '' ? '' : 'Invalid Password');
--- a/src/app/(auth)/signup/page.tsx
+++ b/src/app/(auth)/signup/page.tsx
@@ -1,7 +1,6 @@
 'use client';
 
 import { useState } from 'react';
-// import { useRouter } from 'next/navigation';
 import isEmail from 'validator/lib/isEmail';
 import TextInput from '@/components/TextInput/index';
 import { H1, H2, H4, LinkColored, P } from '@/styles/text';
@@ -10,19 +9,21 @@ import COLORS from '@/styles/colors';
 import { SpacerDiv, HorizontalDiv, H4Centered } from '@/app/(auth)/styles';
 import BigButton from '@/components/BigButton';
 import Button from '@/components/Button';
+import { useAuth } from '@/utils/AuthProvider';
 
 export default function SignUp() {
+  const auth = useAuth();
   const [email, setEmail] = useState('');
   const [password, setPassword] = useState('');
   const [emailSentCount, setEmailSentCount] = useState(0);
   const [errorMessage, setErrorMessage] = useState('');
   const [emailError, setEmailError] = useState('');
   const [passwordError, setPasswordError] = useState('');
-  // const { push } = useRouter();
 
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
   const handleSignUp = async () => {
+    if (!auth) return;
     setEmailError(validEmail(email) ? '' : 'Invalid Email');
     setPasswordError(password !== '' ? '' : 'Invalid Password');
     if (!validEmail(email) || password === '') {
@@ -31,13 +32,9 @@ export default function SignUp() {
     }
     setEmailError('');
     setPasswordError('');
-    const { error } = await supabase.auth.signUp({
-      email,
-      password,
-      options: {
-        emailRedirectTo:
-          'https://immigration-justice-project.vercel.app/email-verified',
-      },
+    const { error } = await auth.signUp(email, password, {
+      emailRedirectTo:
+        'https://immigration-justice-project.vercel.app/email-verified',
     });
 
     if (error) {
--- a/src/app/settings/page.tsx
+++ b/src/app/settings/page.tsx
@@ -8,7 +8,6 @@ import React, {
   useState,
 } from 'react';
 import { useRouter } from 'next/navigation';
-import supabase from '@/api/supabase/createClient';
 import { BackLink, H1, H4 } from '@/styles/text';
 import Button, { LinkButton } from '@/components/Button';
 import COLORS from '@/styles/colors';
@@ -19,6 +18,7 @@ import SettingsSection from '@/components/SettingsSection';
 import { ProfileContext } from '@/utils/ProfileProvider';
 import { isValidBarNumber } from '@/utils/helpers';
 import { ButtonContainer, ContentContainer, PageContainer } from './styles';
+import { useAuth } from '@/utils/AuthProvider';
 
 const rolesOptions = new Map<RoleEnum, string>([
   ['ATTORNEY', 'Attorney'],
@@ -28,11 +28,12 @@ const rolesOptions = new Map<RoleEnum, string>([
 const eoirRegisteredOptions = ['Yes', 'No'];
 
 export default function Settings() {
+  const auth = useAuth();
   const { push } = useRouter();
   const profile = useContext(ProfileContext);
 
-  const userId = useMemo(() => profile?.userId, [profile]);
-  const userEmail = useMemo(() => profile?.userEmail, [profile]);
+  const userId = useMemo(() => auth?.userId, [profile]);
+  const userEmail = useMemo(() => auth?.userEmail, [profile]);
 
   const [basicInformation, setBasicInformation] = useState<SettingsSectionData>(
     [],
@@ -167,7 +168,7 @@ export default function Settings() {
   }, [profile]);
 
   const handleSignOut = useCallback(async () => {
-    const { error } = await supabase.auth.signOut();
+    const error = await auth?.signOut();
     if (error) {
       throw new Error(`An error occurred trying to sign out: ${error.message}`);
     }
--- a/src/app/settings/page.tsx
+++ b/src/app/settings/page.tsx
@@ -1,12 +1,6 @@
 'use client';
 
-import React, {
-  useCallback,
-  useContext,
-  useEffect,
-  useMemo,
-  useState,
-} from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useRouter } from 'next/navigation';
 import { BackLink, H1, H4 } from '@/styles/text';
 import Button, { LinkButton } from '@/components/Button';
@@ -15,10 +9,10 @@ import { cities, languages } from '@/lib/bigData';
 import { ProfileLanguage, ProfileRole, RoleEnum } from '@/types/schema';
 import { SettingsSectionData, SubSectionData } from '@/types/settingsSection';
 import SettingsSection from '@/components/SettingsSection';
-import { ProfileContext } from '@/utils/ProfileProvider';
+import { useProfile } from '@/utils/ProfileProvider';
 import { isValidBarNumber } from '@/utils/helpers';
-import { ButtonContainer, ContentContainer, PageContainer } from './styles';
 import { useAuth } from '@/utils/AuthProvider';
+import { ButtonContainer, ContentContainer, PageContainer } from './styles';
 
 const rolesOptions = new Map<RoleEnum, string>([
   ['ATTORNEY', 'Attorney'],
@@ -30,10 +24,10 @@ const eoirRegisteredOptions = ['Yes', 'No'];
 export default function Settings() {
   const auth = useAuth();
   const { push } = useRouter();
-  const profile = useContext(ProfileContext);
+  const profile = useProfile();
 
-  const userId = useMemo(() => auth?.userId, [profile]);
-  const userEmail = useMemo(() => auth?.userEmail, [profile]);
+  const userId = useMemo(() => auth?.userId, [auth]);
+  const userEmail = useMemo(() => auth?.userEmail, [auth]);
 
   const [basicInformation, setBasicInformation] = useState<SettingsSectionData>(
     [],
--- a/src/app/settings/page.tsx
+++ b/src/app/settings/page.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useRouter } from 'next/navigation';
 import { BackLink, H1, H4 } from '@/styles/text';
 import Button, { LinkButton } from '@/components/Button';
@@ -167,7 +167,7 @@ export default function Settings() {
       throw new Error(`An error occurred trying to sign out: ${error.message}`);
     }
     push('/login');
-  }, [push]);
+  }, [auth, push]);
 
   const handleUpdateBasicInformation = useCallback(
     (nd: SettingsSectionData) => {
--- a/src/components/CaseDetails/index.tsx
+++ b/src/components/CaseDetails/index.tsx
@@ -4,7 +4,6 @@ import {
   timestampStringToDate,
   parseDate,
   parseAgency,
-  parseExperience,
   parseRolesNeeded,
   parseTimeCommitment,
 } from '@/utils/helpers';
@@ -71,11 +70,6 @@ const caseFields = [
     getValue: (data: CaseListing) => data.languages.join(', '),
   },
   {
-    label: 'Attorney Experience Level',
-    getValue: (data: CaseListing) =>
-      parseExperience(data.experience_needed) || 'N/A',
-  },
-  {
     label: 'Relief Sought',
     getValue: (data: CaseListing) => data.relief_codes.join(', ') || 'N/A',
   },
@@ -93,8 +87,8 @@ export default function CaseDetails({ caseData }: { caseData: CaseListing }) {
   const profile = useProfile();
 
   const Interest = useMemo(() => {
-    if (auth && auth.userId && profile && profile.profileData)
-      return profile.profileData ? (
+    if (auth && auth.userId)
+      return profile?.profileData ? (
         <InterestForm caseData={caseData} />
       ) : (
         <>
--- a/src/components/CaseDetails/index.tsx
+++ b/src/components/CaseDetails/index.tsx
@@ -134,7 +134,7 @@ export default function CaseDetails({ caseData }: { caseData: CaseListing }) {
         </AuthButtons>
       </>
     );
-  }, [caseData, profile]);
+  }, [auth, caseData, profile]);
 
   return (
     <CaseDisplay>
--- a/src/components/CaseDetails/index.tsx
+++ b/src/components/CaseDetails/index.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-no-comment-textnodes */
-import React, { useContext, useMemo } from 'react';
+import React, { useMemo } from 'react';
 import {
   timestampStringToDate,
   parseDate,
--- a/src/components/CaseDetails/index.tsx
+++ b/src/components/CaseDetails/index.tsx
@@ -4,6 +4,7 @@ import {
   timestampStringToDate,
   parseDate,
   parseAgency,
+  //parseExperience,
   parseRolesNeeded,
   parseTimeCommitment,
 } from '@/utils/helpers';
@@ -69,6 +70,11 @@ const caseFields = [
     label: 'Client Language',
     getValue: (data: CaseListing) => data.languages.join(', '),
   },
+  // {
+  //   label: 'Attorney Experience Level',
+  //   getValue: (data: CaseListing) =>
+  //     parseExperience(data.experience_needed) || 'N/A',
+  // },
   {
     label: 'Relief Sought',
     getValue: (data: CaseListing) => data.relief_codes.join(', ') || 'N/A',
--- a/src/components/CaseDetails/index.tsx
+++ b/src/components/CaseDetails/index.tsx
@@ -4,7 +4,7 @@ import {
   timestampStringToDate,
   parseDate,
   parseAgency,
-  //parseExperience,
+  parseExperience,
   parseRolesNeeded,
   parseTimeCommitment,
 } from '@/utils/helpers';
@@ -70,11 +70,11 @@ const caseFields = [
     label: 'Client Language',
     getValue: (data: CaseListing) => data.languages.join(', '),
   },
-  // {
-  //   label: 'Attorney Experience Level',
-  //   getValue: (data: CaseListing) =>
-  //     parseExperience(data.experience_needed) || 'N/A',
-  // },
+  {
+    label: 'Attorney Experience Level',
+    getValue: (data: CaseListing) =>
+      parseExperience(data.experience_needed) || 'N/A',
+  },
   {
     label: 'Relief Sought',
     getValue: (data: CaseListing) => data.relief_codes.join(', ') || 'N/A',
--- a/src/components/CaseDetails/index.tsx
+++ b/src/components/CaseDetails/index.tsx
@@ -9,7 +9,8 @@ import {
 } from '@/utils/helpers';
 import { H2, H3, P, StrongP } from '@/styles/text';
 import { CaseListing } from '@/types/schema';
-import { ProfileContext } from '@/utils/ProfileProvider';
+import { useProfile } from '@/utils/ProfileProvider';
+import { useAuth } from '@/utils/AuthProvider';
 import COLORS from '@/styles/colors';
 import InterestForm from '../InterestForm';
 import { LinkButton } from '../Button';
@@ -82,10 +83,11 @@ const caseFields = [
 ];
 
 export default function CaseDetails({ caseData }: { caseData: CaseListing }) {
-  const profile = useContext(ProfileContext);
+  const auth = useAuth();
+  const profile = useProfile();
 
   const Interest = useMemo(() => {
-    if (profile && profile.userId)
+    if (auth && auth.userId && profile)
       return profile.profileData ? (
         <InterestForm caseData={caseData} />
       ) : (
--- a/src/components/CaseDetails/index.tsx
+++ b/src/components/CaseDetails/index.tsx
@@ -93,7 +93,7 @@ export default function CaseDetails({ caseData }: { caseData: CaseListing }) {
   const profile = useProfile();
 
   const Interest = useMemo(() => {
-    if (auth && auth.userId && profile)
+    if (auth && auth.userId && profile && profile.profileData)
       return profile.profileData ? (
         <InterestForm caseData={caseData} />
       ) : (
--- a/src/components/InterestForm/index.tsx
+++ b/src/components/InterestForm/index.tsx
@@ -29,12 +29,13 @@ const radioOptions = [
 ];
 
 export default function InterestForm({ caseData }: { caseData: CaseListing }) {
+  const auth = useAuth();
+  const profile = useProfile();
   const [reason, setReason] = useState<string>('');
   const [rolesInterested, setRolesInterested] = useState<string>('');
   const [startDate, setStartDate] = useState<string>('');
   const [submitted, setSubmitted] = useState(false);
   const [missingInfo, setMissingInfo] = useState(false);
-  const profile = useContext(ProfileContext);
 
   useEffect(() => {
     // Reset form fields when caseData changes
@@ -51,11 +52,11 @@ export default function InterestForm({ caseData }: { caseData: CaseListing }) {
       return;
     }
     if (isValidDate(startDate)) {
-      if (profile && profile.userId) {
+      if (auth && auth.userId && profile) {
         const newInterest: Interest = {
           listing_id: caseData.id,
           listing_type: 'CASE',
-          user_id: profile.userId,
+          user_id: auth.userId,
           form_response: {
             start_date: new Date(startDate),
             interest_reason: reason,
--- a/src/components/InterestForm/index.tsx
+++ b/src/components/InterestForm/index.tsx
@@ -1,9 +1,9 @@
 'use client';
 
-import { useState, useEffect, useContext } from 'react';
+import { useState, useEffect } from 'react';
 import { upsertInterest } from '@/api/supabase/queries/interest';
-import { Interest, CaseListing } from '@/types/schema';
-import { ProfileContext } from '@/utils/ProfileProvider';
+import { Interest, CaseListing, RoleEnum } from '@/types/schema';
+import { useAuth } from '@/utils/AuthProvider';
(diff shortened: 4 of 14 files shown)


src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+      const sessionUserEmail = newSession.user.email;
+      setUserId(sessionUserId);
+      setUserEmail(sessionUserEmail);
+    }
+  };
+
+  // on page load, check if there's a session and set it
+  useEffect(() => {
+    supabase.auth.getSession().then(({ data: { session: newSession } }) => {
+      setAll(newSession != null ? newSession : undefined);
```
jinkang-0: this seems a little redundant, we could just toss `newSession` into the function and handle the `null` case there (since it already handles falsy values)

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }, []);
+
+  // sign in and set the session, userId, and userEmail
+  const signIn = async (email: string, password: string) => {
+    const value = await supabase.auth.signInWithPassword({
+      email,
+      password,
+    }); // will trigger onAuthStateChange to update the session
+
+    setAll(value.data.session ? value.data.session : undefined);
```
jinkang-0: can't we just toss `value.data.session`...

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  // sign out and clear the session, userId, and userEmail
+  const signOut = async () => {
+    const { error } = await supabase.auth.signOut();
+
+    setSession(undefined);
+    setUserId(undefined);
+    setUserEmail(undefined);
+
+    return error;
```
jinkang-0: i wonder what error we'll have - if there's an error, should we still set these to be undefined?

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  // sign up and set the session, userId, and userEmail
+  const signUp = async (email: string, password: string, options: object) => {
+    console.log('signUp', email, password, options);
```
jinkang-0: should we remove this `console.log`?

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    setSession(undefined);
+    setUserId(undefined);
+    setUserEmail(undefined);
+
+    return error;
+  };
+
+  // sign up and set the session, userId, and userEmail
+  const signUp = async (email: string, password: string, options: object) => {
```
jinkang-0: i feel weird having `options` be an `object` type, since it won't provide us with any option suggestions in VS code when we use the function (as opposed to using `supabase.auth.signUp`.

It would be best if we can find a type that corresponds to what we want to pass into `supabase.auth.signUp`, either as a type import from supabase or as a custom defined interface

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+    () => ({
+      session,
+      userId,
+      userEmail,
+      signIn,
+      signOut,
+      signUp,
+    }),
+    [session, userId, userEmail, signIn, signOut, signUp],
+  );
```
jinkang-0, line 119: instead of this, i would recommend wrapping each function inside of a `useCallback` and just put the values into the `AuthContext.Provider` component (i.e. no `useMemo`).

reasoning (mansplaining):
- if **anything** is changed inside a React component, **everything** will be recomputed and re-rendered
- this means that if anything changes, all the functions (which are defined as local variables) will be re-rendered
- as a consequence, everything in the `useMemo` hook will be updated each time the component re-renders, which kind of defeats the whole purpose of the `useMemo` hook
- wrapping each function in a `useCallback` will fix this (and i recommend this), as it tells React to only update the function if anything in the dependency array of the `useCallback` hook is changed
    - in other words, the `useMemo` hook is no longer required
- an alternative is to define the functions inside of `useMemo` (as i have in the other contexts), but this is less efficient, since it will redefine **all** functions regardless of their respective dependencies each time **any** value updates

jinkang-0: i recognize i made the same mistake in `ProfileProvider` and `OnboardingProvider` as well LOL
would be great if you want to update that too 👍

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+    options: object,
+  ) => Promise<AuthResponse>;
+}
+
+export const AuthContext = createContext<AuthContextType | undefined>(
+  undefined,
+);
+
+// TODO: we might need error handling here?
+export const useAuth = () => useContext(AuthContext);
```
jinkang-0: i think error handling should be placed in their respective functions, no need for that here

src/utils/ProfileProvider.tsx
```diff
@@ -47,9 +46,11 @@ export const ProfileContext = createContext<ProfileContextType | undefined>(
   undefined,
 );
 
+// TODO: we might need error handling here?
+export const useProfile = () => useContext(ProfileContext);
```
jinkang-0: similar: ifl error handling should be handled in their respective functions

src/utils/ProfileProvider.tsx
```diff
@@ -58,31 +59,16 @@ export default function ProfileProvider({ children }: { children: ReactNode }) {
...
-      fetchLanguagesById(sessionUserId).then(data => setProfileLangs(data)),
-      fetchRolesById(sessionUserId).then(data => setProfileRoles(data)),
+      fetchProfileById(auth?.userId).then(data => setProfileData(data)),
+      fetchLanguagesById(auth?.userId).then(data => setProfileLangs(data)),
+      fetchRolesById(auth?.userId).then(data => setProfileRoles(data)),
     ]);
 
     setProfileReady(true);
-  }, []);
+  }, [auth]); // TODO: should auth be a dependency here?
```
jinkang-0: response to `TODO`: yep auth should be a dependency, i would recommend adding the ESLint extension on VS Code

no issues here

src/app/(auth)/signup/page.tsx
```diff
@@ -10,19 +9,22 @@ import COLORS from '@/styles/colors';
...
   const [emailSentCount, setEmailSentCount] = useState(0);
   const [errorMessage, setErrorMessage] = useState('');
   const [emailError, setEmailError] = useState('');
   const [passwordError, setPasswordError] = useState('');
-  // const { push } = useRouter();
 
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
   const handleSignUp = async () => {
+    console.log(auth);
```
jinkang-0: we should probably remove this `console.log`

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  // sign up and set the session, userId, and userEmail
+  const signUp = async (email: string, password: string, options: object) => {
+    console.log('signUp', email, password, options);
+    const value = await supabase.auth.signUp({
+      email,
+      password,
+      options: {
+        ...options,
+      },
```
jinkang-0: can't we just... pass in `options`...

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }); // will trigger onAuthStateChange to update the session
+
+    setAll(value.data.session ? value.data.session : undefined);
```
jinkang-0: flagging here as well: i think we can just toss `value.data.session`

src/components/CaseDetails/index.tsx
```diff
@@ -88,10 +89,11 @@ const caseFields = [
...
-  const profile = useContext(ProfileContext);
+  const auth = useAuth();
+  const profile = useProfile();
 
   const Interest = useMemo(() => {
-    if (profile && profile.userId)
+    if (auth && auth.userId && profile)
       return profile.profileData ? (
         <InterestForm caseData={caseData} />
       ) : (
```
jinkang-0: we can remove the `profile` check, since the values of the context is not used at all

additionally, can you replace the `profile` with `auth` in the dependency array?

src/app/settings/page.tsx
```diff
@@ -194,7 +189,7 @@ export default function Settings() {
   }, [profile]);
 
   const handleSignOut = useCallback(async () => {
-    const { error } = await supabase.auth.signOut();
+    const error = await auth?.signOut();
     if (error) {
       throw new Error(`An error occurred trying to sign out: ${error.message}`);
     }
```
jinkang-0, line 168: can you add `auth` to the dependency array of this `useCallback` hook?

src/app/settings/page.tsx
```diff
@@ -1,14 +1,7 @@
 'use client';
 
-import React, {
-  useCallback,
-  useContext,
-  useEffect,
-  useMemo,
-  useState,
-} from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
jinkang-0: this is my bad, but we could lowk remove `React` from here, don't think we access that anywhere here lol

src/components/InterestForm/index.tsx
```diff
@@ -3,7 +3,8 @@
 import { useState, useEffect, useContext } from 'react';
```
jinkang-0: we're no longer using `useContext`, so we can remove that import

src/components/InterestForm/index.tsx
```diff
@@ -51,11 +53,11 @@ export default function InterestForm({ caseData }: { caseData: CaseListing }) {
       return;
     }
     if (isValidDate(startDate)) {
-      if (profile && profile.userId) {
+      if (auth && auth.userId && profile) {
```
jinkang-0: we don't need to use the `profile` context here, so we can remove this check

src/components/InterestForm/index.tsx
```diff
@@ -29,12 +30,13 @@ const radioOptions = [
 ];
 
 export default function InterestForm({ caseData }: { caseData: CaseListing }) {
+  const auth = useAuth();
+  const profile = useProfile();
```
jinkang-0: we won't need to use the profile context here, so we can delete this line

src/components/InterestForm/index.tsx
```diff
@@ -3,7 +3,8 @@
 import { useState, useEffect, useContext } from 'react';
 import { upsertInterest } from '@/api/supabase/queries/interest';
 import { Interest, CaseListing, RoleEnum } from '@/types/schema';
-import { ProfileContext } from '@/utils/ProfileProvider';
+import { useProfile } from '@/utils/ProfileProvider';
```
jinkang-0: since we won't be using the profile context, we don't need to import this hook

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
+'use client';
+
+import { AuthError, AuthResponse, Session } from '@supabase/supabase-js';
+import { createContext, useContext, useEffect, useMemo, useState } from 'react';
+import { UUID } from 'crypto';
+import supabase from '@/api/supabase/createClient';
```
jinkang-0: small nit: we can use `import type` for imports from `@supabase/supabase-js` and `crypto`, since we'll only need the type declarations

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+  userId: UUID | undefined;
+  userEmail: string | undefined;
+  signIn: (email: string, password: string) => Promise<AuthResponse>;
+  signOut: () => Promise<AuthError | null>;
+  signUp: (
+    email: string,
+    password: string,
+    options: object,
+  ) => Promise<AuthResponse>;
+}
```
jinkang-0: regarding the `TODO` - not sure how much it matters whether we're consistent with `null` or `undefined`, but I think there could be a subtle difference that can make things cleaner

for me, i see it as:
- `undefined` = the value haven't been initialized yet
- `null` = the value doesn't exist at all

ngl i think their difference is negligible and we can just use truthy value to differentiate those from usable info, but i'm curious to hear your thoughts on this

src/utils/OnboardingProvider.tsx
```diff
@@ -15,8 +9,16 @@ import {
   Dispatch,
   useContext,
```
jinkang-0: we're no longer using `useContext`, so we can remove that import

src/utils/OnboardingProvider.tsx
```diff
@@ -76,11 +79,12 @@ export default function OnboardingProvider({
...
 
     const flushData = async () => {
+      if (!auth) throw new Error('Fatal: No auth context provided!');
       if (!profileCtx) throw new Error('Fatal: No profile context provided!');
-      if (profileCtx.userId === undefined)
+      if (auth.userId === undefined)
         throw new Error('Fatal: User is not logged in!');
 
-      const uid: UUID = profileCtx.userId;
+      const uid: UUID = auth.userId;
```
jinkang-0: can you add `auth` to either:
- the dependency array of the `useMemo` hook; or
- if you decide to implement the `useCallback` hooks here, then the dependency array of that

src/utils/OnboardingProvider.tsx
```diff
@@ -50,7 +52,8 @@ export default function OnboardingProvider({
 }: {
   children: ReactNode;
 }) {
-  const profileCtx = useContext(ProfileContext);
+  const auth = useAuth();
+  const profileCtx = useProfile();
```
jinkang-0: microscopic nit: should we standardize to call it `profile` vs `profileCtx`? not sure how much this matters. fine with either. 

side tip that im sure you know already: if we choose to refactor, you can put the text cursor on `profileCtx` then hit F2 when the keyword is highlighted to rename all instances of `profileCtx`

src/utils/OnboardingProvider.tsx
```diff
@@ -76,11 +79,12 @@ export default function OnboardingProvider({
     };
 
     const flushData = async () => {
+      if (!auth) throw new Error('Fatal: No auth context provided!');
       if (!profileCtx) throw new Error('Fatal: No profile context provided!');
-      if (profileCtx.userId === undefined)
+      if (auth.userId === undefined)
```
jinkang-0: i would say it's better to use `!auth.userId` here instead (idk what i was thinking before) since it would check for all of:
1. `''`
2. `null`
3. `undefined`

src/utils/ProfileProvider.tsx
```diff
@@ -100,7 +86,7 @@ export default function ProfileProvider({ children }: { children: ReactNode }) {
       newLanguages: ProfileLanguage[],
       newRoles: ProfileRole[],
     ) => {
-      if (!userId) throw new Error('Fatal: user is not logged in!');
+      if (!auth?.userId) throw new Error('Fatal: user is not logged in!');
       if (profileData) throw new Error('Fatal: user already has a profile!');
```
jinkang-0: i just realized we're checking for `auth.userId` but not using it here...
should we add a check for `auth.userId === newProfile.user_id`?

src/utils/OnboardingProvider.tsx
varortz: nit: I also prefer `profile` vs `profileCtx` lol

src/utils/AuthProvider.tsx
varortz: looked more into this and there are a lot of `null` haters, specifically for javascript/typescript. Also apparently Typescrip Coding Guidelines advise against the use of `null`, and Tony Hoare—the creator of null—considers it a billion dollar mistake 💀. Here are some articles I found:
- [We don't need null ](https://lou.cx/articles/we-don-t-need-null/)
- [null vs undefined? What to choose? What to use?](https://dev.to/sduduzog/null-vs-undefined-what-to-choose-what-to-use-11g)

so maybe `undefined` is the way to go for this use case?

jinkang-0: I don't think we should go full on `null` or full on `undefined`. They both have their own use cases, as the second article you provided described (thanks for doing the research I couldn't be bothered to do). I would be weary of following those practices to a T, especially if Typescript already gives us [a way to detect both at once](https://medium.com/@teamcode20233/understanding-double-question-mark-operator-in-typescript-be869f210fe6) (although this is more syntactic sugar).

I think, for the use case here, it makes sense for `signOut` to return `null` if there is no error. If we used `undefined`, it would sort of be saying that there was an error in getting the error, if that makes any sense at all. Or that the error wasn't set, even though it should've.

It also makes sense for `userId` and `userEmail` to be optionally `undefined`, because we may not have fetched that data yet. We can make an argument for `null` if the field can be fetched and is empty, but I don't think we'll encounter that error at least for `userId` and `userEmail`.

`session` is interesting though, since it can be `null` or `undefined`. In this case, I would go with defaulting to `null` just to have an empty value.

That being said, I do think we should be consistent across the app about our usage of `null` and `undefined`. I've probably made the same mistake about a hundred times.

If it makes it easier for everyone (since lowk I won't care what `null` or `undefined` means), we can go with `undefined` (borrowing [this philosophy](https://www.reddit.com/r/typescript/comments/11dpu05/comment/jaagvta/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)). This feels like taking the lesser evil to me though, and I still think there could be merit in encoding different semantic meanings in `undefined` and `null`.

jinkang-0: side note: we can reduce the definition to be `session?: Session`, which is Typescript's way of writing the same thing.

varortz: yeah, that makes sense. It does seem like there are cases where `null` makes sense over `undefined`, and we should probably try and be consistent across the app as you mentioned

src/utils/AuthProvider.tsx
pragyakallanagoudar: is this the same concern as the one above? i.e. "we could just toss `newSession` into the function and handle the null case there" -- just want to make sure i'm catching everything

src/utils/AuthProvider.tsx
pragyakallanagoudar: hmmm yeah i have no idea -- seems like the only one on [here](https://supabase.com/docs/reference/javascript/auth-signout) is you can't sign out if you're not already signed in, which shouldn't be possible anyways because of the way this is designed (i.e. sign out button doesn't show up if you're not logged in)

i think gwn does not check for the sign out error? not sure if we can just skip it

but yes im changing it so it only sets things to undefined if there is no error

src/utils/AuthProvider.tsx
pragyakallanagoudar: are there situations where we will be calling `signUp` outside of the current instance in the signup page, though? this feels like a concern only if we were to call the function in different instances in the future, but it's not clear to me where this might be. maybe to make this cleaner, i can get rid of the `options` arg entirely since idk that it has much utility lol

src/utils/ProfileProvider.tsx
pragyakallanagoudar: downloaded it, it better be life-altering

src/utils/AuthProvider.tsx
pragyakallanagoudar: thanks jinkang!! i think i'm a little confused about the syntax of the fix lol so let's discuss on sunday

src/utils/OnboardingProvider.tsx
pragyakallanagoudar: i think the reason this is `profileCtx` here is because there is also this `useState` defined a few lines later:

`const [profile, setProfile] = useState<Partial<Profile>>({});`

shall i rename this instead? and if so, what to? maybe `onboardingProfile` or `profileObj` or `profileState` or smth?

src/components/CaseDetails/index.tsx
pragyakallanagoudar: hmm can you clarify how they're not being used? aren't they relevant for the un-onboarded state

src/components/InterestForm/index.tsx
pragyakallanagoudar: for this and all the other comments about getting rid of the `ProfileContext` in the `InterestForm` -

@varortz I was thinking the `ProfileContext` would be relevant to the profile match -- should I leave this here for that reason?

src/utils/OnboardingProvider.tsx
jinkang-0: oh mb i overlooked that

maybe we can call it `userProfile`?

src/components/CaseDetails/index.tsx
jinkang-0: oh i overlooked the context.

currently you're only checking if the `profile` object exists, which represents the profile context. But that doesn't tell us anything about the actual onboarded-ness of the user.

taking into account the context of the use case, a better suggestion might be to use `profile.profileData != null` instead

src/utils/ProfileProvider.tsx
jinkang-0: it'll be just like brookies

src/utils/AuthProvider.tsx
jinkang-0: yes

src/utils/AuthProvider.tsx
jinkang-0: ohh that makes sense

on a slightly less relevant tangent --
looking at the docs, i noticed this blurb:
> By default, `signOut()` uses the global scope, which signs out all other sessions that the user is logged into as well.

given this, should we add a check in `onAuthStateChange` to set those fields to be `undefined` when we receive a log out event? In fact, should we just move all our state changes into the `onAuthStateChange` function depending on the event received?

not sure if that would work, but it may make things a lot cleaner and secure

src/utils/AuthProvider.tsx
jinkang-0: I doubt it, but it's still good practice to have well-defined types. I can imagine a scenario where we may want to modify arguments to the `signUp` function but get mildly inconvenienced by the lack of IntelliSense. Which is the worst inconvenience to have if I'm being completely honest.

Anywho, if it helps, a quick dive into the node modules give us this declaration:
```Typescript
options?: {
  emailRedirectTo?: string;
  data?: object;
  captchaToken?: string;
}
```

But I'll leave the choice (and blame) of implementing this up to you.

note that `options` is optional tho (can be `undefined`)

src/components/InterestForm/index.tsx
varortz: I think profile match would be its own separate component, so it should be ok to remove it in the `InterestForm` if we're not using it

src/utils/OnboardingProvider.tsx
pragyakallanagoudar: hmmm `userProfile` (and `userState`, my other idea) also don't work because there are types imported from `@/types/schema` that have the same name. imma use `profileObj` lol

jinkang-0: Can you elaborate on where they're imported? I don't see `userProfile` or `userState` being defined (nor exported) from `@/types/schema`
image
image

jinkang-0: but `profileObj` would work too

src/utils/AuthProvider.tsx
jinkang-0: In hindsight, I doubt we'd need to use a `useCallback` hook at all, since they don't return a value for the memoization to be helpful. We could probably just define the functions as we would for normal handlers and toss that along the other values (which may or may not be wrapped in `useMemo`).

src/utils/OnboardingProvider.tsx
pragyakallanagoudar: i actually. have literally no clue what i was referring to LOL. `userProfile` worked! replacing

src/utils/OnboardingProvider.tsx
pragyakallanagoudar: added it to the dependency array of the useMemo hook!

src/utils/AuthProvider.tsx
pragyakallanagoudar: (Recap of what we discussed this during MPH) It seems like using this causes a TS compiler error, so I'm just using what we already had for right now. but @jinkang-0 @varortz curious if you have other approaches you'd like to suggest?

pragyakallanagoudar: Update: I wrapped the auth functions in `useCallback`s and left everything else as is, and this seems to get rid of all TS compiler errors -- but let me know what y'all think!!

src/utils/AuthProvider.tsx
pragyakallanagoudar: Sounds good! I've replaced it with `session?: Session`

src/utils/AuthProvider.tsx
pragyakallanagoudar: @varortz thoughts on moving things to `onAuthStateChange`? im chill w anything

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const setAll = (newSession: Session | null) => {
+    if (!newSession) return;
+    setSession(newSession);
+    if (newSession?.user?.id) {
+      const sessionUserId = newSession.user.id as UUID;
+      const sessionUserEmail = newSession.user.email;
+      setUserId(sessionUserId);
+      setUserEmail(sessionUserEmail);
+    }
+  };
```
jinkang-0: nit: we can wrap this in a `useCallback` as well

src/utils/OnboardingProvider.tsx
```diff
@@ -71,49 +73,49 @@ export default function OnboardingProvider({
      * Does not affect database.
      */
     const updateProfile = (updatedInfo: Partial<Profile>) => {
-      const newProfileData = { ...profile, ...updatedInfo };
-      setProfile(newProfileData);
+      const newUserProfile = { ...userProfile, ...updatedInfo };
+      setUserProfile(newUserProfile);
     };
```
jinkang-0: for this and all other functions in the `useMemo`, could you put them outside the `useMemo` and wrap each of them individually in a `useCallback`?

with the current configuration, all of these functions will be re-defined every time the `useMemo` is re-computed (which happens a lot) - ideally we'd just want to define functions once.

src/utils/ProfileProvider.tsx
```diff
@@ -58,31 +58,16 @@ export default function ProfileProvider({ children }: { children: ReactNode }) {
...
-    setUserId(sessionUserId);
+    if (!auth?.userId) return;
 
     await Promise.all([
-      fetchProfileById(sessionUserId).then(data => setProfileData(data)),
-      fetchLanguagesById(sessionUserId).then(data => setProfileLangs(data)),
-      fetchRolesById(sessionUserId).then(data => setProfileRoles(data)),
+      fetchProfileById(auth?.userId).then(data => setProfileData(data)),
+      fetchLanguagesById(auth?.userId).then(data => setProfileLangs(data)),
+      fetchRolesById(auth?.userId).then(data => setProfileRoles(data)),
```
jinkang-0: nit: correct me if I'm wrong, but since we check for `auth?.userId` (which checks `auth` and `auth.userId`), I think we can just drop the question marks here

src/utils/ProfileProvider.tsx
```diff
@@ -117,10 +104,12 @@ export default function ProfileProvider({ children }: { children: ReactNode }) {
      * setup for the profiles table.
      */
     const updateProfile = async (updatedInfo: Partial<Profile>) => {
```
jinkang-0: can you do something similar with the `useCallback`s here? this is completely my fault - sorry about this.

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+  useMemo,
+  useState,
+} from 'react';
+import type { UUID } from 'crypto';
+import supabase from '@/api/supabase/createClient';
+
+export interface AuthContextType {
+  session?: Session;
+  userId: UUID | undefined;
+  userEmail: string | undefined;
```
jinkang-0: nit: we can use the optional marker here (i.e. `userId?: UUID`)

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,126 @@
...
+  signIn: (email: string, password: string) => Promise<AuthResponse>;
+  signOut: () => Promise<AuthError | null>;
+  signUp: (
+    email: string,
+    password: string,
+    options: object,
```
jinkang-0, line 24: nit: let's use `SignUpOptions` here

jinkang-0: also thoughts on consolidating these miscellaneous types into a `@/types/misc.d.ts` file? (cc @varortz )

jinkang-0 (approved):
amazing work pragya! should be good to go after rebasing and iterations, I'll approve now to speed things up a bit, but feel free to ping me if you want me to take another look

src/utils/AuthProvider.tsx
varortz: would this also include types like `OnboardingContextType`, `ProfileContextType`?

jinkang-0: No, I think context types are best left in the relevant context files. In hindsight, it may be better to leave the `SignUpOptions` type here since we don't need to reference the type anywhere else as well.

varortz: Yeah, if it isn't used somewhere else, it's probably best to leave it in this file.

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,128 @@
...
+}: {
+  children: React.ReactNode;
+}) {
+  const [session, setSession] = useState<Session | undefined>(undefined);
+  const [userId, setUserId] = useState<UUID | undefined>(undefined);
+  const [userEmail, setUserEmail] = useState<string | undefined>(undefined);
+
+  // set the session, userId, and userEmail
+  const setAll = (newSession: Session | null) => {
+    console.log(newSession);
```
jinkang-0: can we remove this console log?

src/utils/AuthProvider.tsx
```diff
@@ -0,0 +1,128 @@
...
+    });
+
+    supabase.auth.onAuthStateChange((_event, newSession) => {
+      setAll(newSession);
+    });
+  }, []);
+
+  // sign in and set the session, userId, and userEmail
+  const signIn = useCallback(async (email: string, password: string) => {
+    console.log('signIn function');
```
jinkang-0: can we remove this console log?

varortz (approved):
looks great!! Thank u so much for all ur hard work :))

PR #65, opened 2024-03-05
[feat] implements the navbar

kevin3656:

- Implemented Nav Bar Component in components. icon redirects to homepage. cases, limited cases and language support redirect to cases page for now. 
- changed authbuttons implementation from cases page to navbar.
- added ijp logo to icon in assets
- Added NavBar Component to cases page, settings page and auth layout.tsx

image

From the homepage you should be able to navigate across pages using nav bar, when logged in nav bar should change to show profile instead of log in/sign up buttons
can check implementation on src/components/NavBar

## :seedling: Next steps 

- update hover effect on links
- fix backlink on settings page(currently overlaps with navbar)

https://www.youtube.com/watch?v=SLfhMt5OUPI

CC: @varortz


src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  text-decoration: none;
+  cursor: pointer;
+
+  &:hover,
+  &:focus {
+    text-decoration: underline;
+  }
+`;
+
+// export const LinkToPage = styled.a`
```
varortz: let's remove these comments if this code is not being used

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  gap: 18px;
+  margin-left: 34px;
+  margin-right: 34px;
+  text-decoration: none;
+`;
+export const AuthButtons = styled.div`
+  display: flex;
+  gap: 1.5625rem;
+  align-self: center;
+  //position: absolute;
```
varortz: let's remove these commented out lines of code, we usually want to avoid leaving in commented code

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  z-index: 1000;
+  width: 100%;
+  box-shadow: 0px 4px 7px 0px rgba(0, 0, 0, 0.1);
+`;
+export const NavBarDiv = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  /* padding-top: 15px;
```
varortz: can we remove these two commented out lines?

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  padding: 15px 0px;
+  justify-items: flex-start;
+  align-items: center;
+`;
+export const IconLink = styled.div`
+  display: flex;
+  gap: 0.5rem;
+`;
+
+export const CustomDiv = styled.div`
```
varortz: could we perhaps change the name `CustomDiv` to something a bit more descriptive to make it more maintainable?

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  display: flex;
+  gap: 1.5625rem;
+  align-self: center;
+  //position: absolute;
+  top: 0;
+  right: 0;
+  //margin-top: 2.1875rem;
+  margin-right: 3.375rem;
+`;
+export const NoUnderlineLink = styled(LinkColored)`
```
varortz: could we update the name `NoUnderlineLink` to something like `NavbarLink`?

src/app/cases/page.tsx
```diff
@@ -153,89 +125,92 @@ export default function Page() {
...
+              placeholder={defaultFilterValues.countries}
+              multi
+              options={countryOptions}
+              value={caseFilters.countries}
+              onChange={v => setCaseFilters({ ...caseFilters, countries: v })}
+            />
+            <ResetFilters onClick={() => resetFilters()}>
+              Reset Filters
+            </ResetFilters>
+            {/* <AuthButtons>{AuthButtonView}</AuthButtons> */}
```
varortz: since we no longer need the `AuthButtons` here anymore, can we just remove this line?

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  /* padding-top: 15px;
+  padding-bottom: 15px; */
+  padding: 15px 0px;
+  justify-items: flex-start;
+  align-items: center;
+`;
+export const IconLink = styled.div`
```
varortz: We can probably remove this `IconLink` div from this file and the `page.tsx` file. `display: flex` is used to create a flex container, and we typically add additional flex properties to adjust the spacing/alignment of its children elements. However, in this specific instance, it appears that there's only one child nested within this div, so we could remove it. Similarly, `gap` would apply to the children elements, so we run into the same situation.

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+`;
+
+export const CustomDiv = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: 18px;
+  margin-left: 34px;
+  margin-right: 34px;
+  text-decoration: none;
```
varortz: I think we can remove this line for `text-decoration` since `CustomDiv` does not have any text itself. Any text nested within this div already handles removing the underline when we aren't hovering over the page names (line 55 - in`NoUnderlineLink`)

src/components/ProfileButton.ts
```diff
@@ -14,7 +13,7 @@ const ProfileButton = styled(Link)`
   background: none;
   border: none;
 
-  color: ${COLORS.greyDarker};
+  color: #ffffff;
   font-size: 1rem;
   font-weight: 600;
   text-decoration: none;
```
varortz: can we update `background-image` in `ProfileButton` so that it includes the updated profile icon (white instead of black icon), so it can match with the designs?

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <>
+        <LinkButton
+          $primaryColor={COLORS.blueMid}
+          $secondaryColor={COLORS.blueDark}
+          href="/login"
+        >
+          Log In
+        </LinkButton>
+        <LinkButton
+          $primaryColor="#CA9B30"
```
varortz: can we add this color to `colors/styles.ts`, maybe we can just call it `goldMid`? We'll likely end up using it in other components later on

src/components/ProfileButton.ts
```diff
@@ -14,7 +13,7 @@ const ProfileButton = styled(Link)`
   background: none;
   border: none;
 
-  color: ${COLORS.greyDarker};
+  color: #ffffff;
```
varortz: can we change `#ffffff` to `white` for readability?

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  /* top: 0;
+  left: 0;
+  position: absolute; */
+  width: 100%;
+  height: 78px;
+  background: ${COLORS.blueMid};
+  align-items: center;
+  z-index: 1000;
+  width: 100%;
+  box-shadow: 0px 4px 7px 0px rgba(0, 0, 0, 0.1);
```
varortz: @kyrenetam do we want box shadow on the navbar? On Figma, the navbar design that was ready for devs did not have one, so I just wanted to make sure since some pages on the design seem to have it

src/app/settings/page.tsx
```diff
@@ -325,73 +326,76 @@ export default function Settings() {
...
-              $primaryColor={COLORS.blueMid}
-              $secondaryColor={COLORS.blueDark}
-              href="/onboarding/roles"
-            >
-              <H4 $color="white">Go to Onboarding</H4>
-            </LinkButton>
+    <>
+      <NavBar />
+      <PageContainer>
+        <BackLink href="/cases">Back</BackLink>
```
varortz: @kyrenetam would we still want a back link on the settings page if we have the navbar?

src/app/(auth)/email-verified/page.tsx
```diff
@@ -4,16 +4,19 @@ import { useRouter } from 'next/navigation';
...
-
   return (
-    <SpacerDiv>
-      <H2>Your email has been verified!</H2>
-      <BigButton type="button" onClick={() => push('/onboarding/roles')}>
-        Go to Onboarding
-      </BigButton>
-    </SpacerDiv>
+    <>
+      <NavBar />
```
ccatherinetan: since the NavBar component was already added in the auth layout, we don't need to add it again on this page! it currently looks like this: 

Screen Shot 2024-03-06 at 9 43 57 PM

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+import COLORS from '@/styles/colors';
+import { openSans } from '@/styles/fonts';
+import { LinkColored } from '@/styles/text';
+
+export const NavBarContainer = styled.div`
+  /* top: 0;
+  left: 0;
+  position: absolute; */
+  width: 100%;
+  height: 78px;
```
ccatherinetan: do we want to change this to rem? i remember @jinkang-0 said that it's best practice to change pixels to rem, especially when something is more than 5 pixels?

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
+import styled from 'styled-components';
+import COLORS from '@/styles/colors';
+import { openSans } from '@/styles/fonts';
+import { LinkColored } from '@/styles/text';
+
+export const NavBarContainer = styled.div`
+  /* top: 0;
```
ccatherinetan: also, just catching another commented line that we may want to delete, going off alvaro!

src/components/NavBar/style.tsx
varortz: over break jinkang told me something about how maybe not everything should be in rem, maybe let's add this to the agenda for tomorrow's worksesh so we're all on the same page with best styling practices!

ccatherinetan (commented):
the nav bar looks rlly good! I couldn't fully review this because the cases page is erroring for me, so i couldn't check the functionality of clicking on cases/language support etc on the navbar... is anyone else having the same problem? (the font also doesn't show up for me? maybe that's a me thing) 

i caught 1 place where the navbar component was erroneously added (in `(auth)/email-verified`). overall, good job! i'll check back in once we figure out the error w/ the cases page T^T 
Screen Shot 2024-03-06 at 9 59 57 PM

src/components/NavBar/style.tsx
kevin3656: Deleted all unnecessary comments .

src/app/(auth)/email-verified/page.tsx
kevin3656: Fixed

src/components/NavBar/style.tsx
kevin3656: Changed to NavBarSectionDiv

src/components/ProfileButton.ts
kevin3656: done

src/components/NavBar/index.tsx
kevin3656: Added goldMid to colors /changed #ffffff to white, changed profile button background-image to white

src/components/NavBar/style.tsx
kevin3656: Done

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+  position: absolute; */
+  width: 100%;
+  height: 78px;
+  background: ${COLORS.blueMid};
+  align-items: center;
+  z-index: 1000;
+  width: 100%;
+  box-shadow: 0px 4px 7px 0px rgba(0, 0, 0, 0.1);
+`;
+export const NavBarDiv = styled.div`
```
varortz: `NavBarDiv` seems a little redundant as it shares a lot of properties with `NavBarContainer`. If you move over the `padding`, `display`, `justify-content`, and `flex-direction` to `NavBarContainer`, we can delete `NavBarDiv` while having the same effect on the navbar.

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { LinkColored } from '@/styles/text';
+
+export const NavBarContainer = styled.div`
+  /* top: 0;
+  left: 0;
+  position: absolute; */
+  width: 100%;
+  height: 78px;
+  background: ${COLORS.blueMid};
+  align-items: center;
```
varortz: we can probably get away with getting rid of `align-items` here since you're already doing this in `NavBarSectionDiv` (previously `CustomDIv`)

src/components/NavBar/style.tsx
varortz: from worksesh: keep drop shadow

src/app/settings/page.tsx
varortz: from worksesh: we can remove it

jinkang-0:
To reduce redundancy and keep merge errors at a minimum, I would recommend placing the `NavBar` component in `app/layout.tsx`.

This may sound like lunacy, but hear me out.

Since the `NavBar` will have to be a client component anyways (to access the `profile` context), it will also be able to access the router, and by extension, the current URL / path of the browser. With this information, we can:
1. Keep the logic of determining which tab is "active" in the component itself (no need to pass props)
2. Determine whether to show or hide the `NavBar` depending on the URL
3. Easily update the logic of the two points above inside the `NavBar` component if we ever decide to change it (without affecting other files)

This may bring a significant change to what you have right now, but I would highly encourage adopting this paradigm as it is used commonly in practice, due to it's ability to take advantage of the `layout` provided by Next.js to avoid re-rendering the `NavBar` each time we visit a new page.

And also make our lives easier. Merge conflicts bad.

src/components/NavBar/style.tsx
```diff
@@ -0,0 +1,44 @@
...
+  display: flex;
+  width: 100%;
+  height: 78px;
+  background: ${COLORS.blueMid};
+  z-index: 1000;
+  width: 100%;
+  box-shadow: 0px 4px 7px 0px rgba(0, 0, 0, 0.1);
+  justify-content: space-between;
+  padding: 15px 0px;
+  position: fixed;
```
varortz: could we actually remove this line for now? We can update this in the second navbar sprint if we want it to be fixed

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { ProfileContext } from '@/utils/ProfileProvider';
+import ProfileButton from '@/components/ProfileButton';
+import { LinkButton } from '@/components/Button';
+import {
+  NavBarContainer,
+  NavBarSectionDiv,
+  AuthButtons,
+  NoUnderlineLink,
+} from './style';
+import Icon from '../../../assets/icons/Icon';
```
ccatherinetan: do we want to change this to `@/assets/icons/Icon` instead?

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+    const currentPath = usePathname();
+    return currentPath.includes(path);
+  }
+
+  type NavLinks = {
+    name: string;
+    path: string;
+    active: boolean;
+  };
+  const navllink: NavLinks[] = [
```
varortz: could we update this const name to `navlink`

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      </>
+    );
+  }, [profile]);
+
+  function IsActive(path: string): boolean {
+    const currentPath = usePathname();
+    return currentPath.includes(path);
+  }
+
+  type NavLinks = {
```
varortz: could we rename `NavLinks` to `NavLink`?

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+          href={link.path}
+          $color="white"
+          isactive={link.active}
+        >
+          {link.name}
+        </Styles.NoUnderlineLink>
+      </div>
+
+      {link.active && (
+        <hr
```
varortz: could we add this into `Navbar/styles.ts` (let's also go ahead and rename the file to `styles.ts` instead of `style.tsx`)

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+    {
+      name: 'Language Support',
+      path: '/language-support',
+      active: IsActive('/language-support'),
+    },
+  ];
+
+  const renderLink = (link: NavLinks) => (
+    <Styles.LinkContainer>
+      <div
```
varortz: I think we can replace this div with `Flex` from `styles/containers.ts`!

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const currentPath = usePathname();
+    return currentPath.includes(path);
+  }
+
+  type NavLink = {
+    name: string;
+    path: string;
+    active: boolean;
+  };
+  const navlink: NavLink[] = [
```
varortz: Ahhh sorry, I realized I forgot the s. I meant to say `navlinks`.

src/components/SmallerButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  font-size: 1rem;
+  font-weight: 500;
+  &:hover {
+    background: ${props => props.$secondaryColor};
+    color: white;
+    border-color: ${props => props.$secondaryColor};
+  }
+`;
+
+const SmallerButton = styled.button`
```
varortz: Let's go ahead and move this to the `Button.tsx` file instead of making this new file.

src/components/SmallerButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  ${ButtonStyles}
+`;
+
+export const LinkButton = styled(Link)`
```
varortz: to avoid redundancy, can we use the `LinkButton` from `Button.tsx`?

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+          Sign Up
+        </LinkButton>
+      </>
+    );
+  }, [profile]);
+
+  function IsActive(path: string): boolean {
+    const currentPath = usePathname();
+    return currentPath.includes(path);
+  }
```
jinkang-0: let's name this `checkIsActive` to follow camel case naming convention for functions, and make it clear that this is a function (using a verb)

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const renderLink = (link: NavLink) => (
+    <Styles.LinkContainer>
+      <Flex
+        style={{
+          display: 'flex',
+          flexDirection: 'column',
+          justifyContent: 'center',
+          flex: 1,
+        }}
+      >
```
jinkang-0: no need to add `display`, `flexDirection`, and `justifyContent` here, we can just use the props for the `Flex` component
`$justify="center"`, `$direction="column"`

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <Styles.NoUnderlineLink
+          href={link.path}
+          $color="white"
+          isactive={link.active}
+        >
+          {link.name}
+        </Styles.NoUnderlineLink>
```
jinkang-0: discussed in final hackathon: change implementation to use a base always bolded text for layout, but toggle visibility with a non-bolded overlay depending on which is active -- this prevents layout changes as we navigate from tab to tab

src/components/NavBar/style.ts
```diff
@@ -0,0 +1,61 @@
...
+  position: relative;
+  height: 100%;
+  margin: 5px;
+  justify-content: space-between;
+  flex-direction: column;
+  align-items: center;
+`;
+
+export const ActiveUnderline = styled.hr<{ isActive: boolean }>`
+  display: ${props => (props.isActive ? 'block' : 'none')};
```
jinkang-0: let's use destructuring here

src/components/NavBar/style.ts
```diff
@@ -0,0 +1,61 @@
...
+  display: flex;
+  position: relative;
+  height: 100%;
+  margin: 5px;
+  justify-content: space-between;
+  flex-direction: column;
+  align-items: center;
+`;
+
+export const ActiveUnderline = styled.hr<{ isActive: boolean }>`
```
jinkang-0: for styled components, props should be prefixed with a `$`, so `$isActive`

src/components/NavBar/style.ts
```diff
@@ -0,0 +1,61 @@
...
+  margin-left: 34px;
+  margin-right: 34px;
+`;
+export const AuthButtons = styled.div`
+  display: flex;
+  gap: 1rem;
+  align-self: center;
+  top: 0;
+  right: 0;
+  margin-right: 3.375rem;
```
jinkang-0: let's remove the margin right here

src/components/NavBar/style.ts
```diff
@@ -0,0 +1,61 @@
...
+`;
+export const NoUnderlineLink = styled(LinkColored)<{ isactive: boolean }>`
+  ${openSans.style}
+  text-decoration: none;
+  margin: 10px;
+  font-weight: ${props => (props.isactive ? '600' : '400')};
```
jinkang-0, line 37: we'll probably need to set this to 600 to ensure the layout-impacting text is always bolded

src/components/Button.tsx
jinkang-0, line 1: is there any reason why we added a separate `Button.tsx` file here instead of incorporating new styles to the existing `Buttons.tsx`?

assets/icons/Icon.tsx
jinkang-0, line 1: let's delete this file - i think we're using `lib/icons.tsx` now!

assets/icons/logo.svg
jinkang-0, line 1: instead of an svg file, let's add the svg definition in `lib/icons.tsx` as a separate icon

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+import { useContext, useMemo } from 'react';
+import Link from 'next/link';
+import { usePathname } from 'next/navigation';
+import ProfileButton from '@/components/ProfileButton';
+import COLORS from '@/styles/colors';
+import { Flex } from '@/styles/containers';
+import { useAuth } from '@/utils/AuthProvider';
+import { useProfile } from '@/utils/ProfileProvider';
+import Icon from '../../../assets/icons/Icon';
```
jinkang-0: I think we will be importing from `components/Icon` now!

src/components/SmallerButton.tsx
jinkang-0, line 1: this seems redundant, we can remove this file after moving the `SmallerButton` logic into `Buttons.tsx`

src/components/NavBar/style.ts
```diff
@@ -0,0 +1,60 @@
...
+  gap: 1rem;
+  align-self: center;
+  top: 0;
+  right: 0;
+`;
+export const NoUnderlineLink = styled(LinkColored)<{ isactive: boolean }>`
+  ${sans.style}
+  text-decoration: none;
+  margin: 10px;
+  font-weight: ${props => (props.isactive ? '600' : '400')};
```
jinkang-0: let's use destructuring here to stay consistent

src/components/NavBar/style.ts
```diff
@@ -0,0 +1,60 @@
...
+`;
+
+export const AuthButtons = styled.div`
+  display: flex;
+  gap: 1rem;
+  align-self: center;
+  top: 0;
+  right: 0;
+`;
+export const NoUnderlineLink = styled(LinkColored)<{ isactive: boolean }>`
```
jinkang-0: can we update the parameter name to `$isActive` to stay consistent with `ActiveUnderline`?

src/components/Button.tsx
```diff
@@ -0,0 +1,89 @@
...
+      props.$primaryColor ? props.$primaryColor : props.$secondaryColor};
+  cursor: pointer;
+  transition: 150ms ease-in-out;
+  font-size: 1rem;
+  font-weight: 500;
+  &:hover {
+    background: ${props => props.$secondaryColor};
+    color: white;
+    border-color: ${props => props.$secondaryColor};
+  }
```
jinkang-0: a lot of this seems to be copied from the `ButtonStyles`
just as a note, we can simply import the styles over like so:
```ts
const SmallButton = styled.button`
  ${ButtonStyles}
  /* styles specific to SmallButton to override regular button styles */
`
```

jinkang-0:
looks good kevin! should be good after addressing some minor nits and prettier complaints!
one note: don't forget to remove the back link on the settings page as well!

src/components/Buttons.tsx
```diff
@@ -101,6 +101,19 @@ export const LinkButton = styled(Link)`
...
 `;
+export const SmallLinkButton = styled(Link)`
+  ${ButtonStyles}
+  text-decoration: none;
+  padding: 0.25rem 0.75rem;
+  font-weight: 400;
+  border-radius: 0.313rem; // 5px
+  border: 2px solid
+    ${props =>
+      props.$primaryColor ? props.$primaryColor : props.$secondaryColor};
```
jinkang-0: can we use destructuring here to stay consistent?

src/components/ProfileButton.ts
```diff
@@ -1,6 +1,5 @@
 import Link from 'next/link';
 import styled from 'styled-components';
-import COLORS from '@/styles/colors';
 import { sans } from '@/styles/fonts';
 
 const ProfileButton = styled(Link)`
```
varortz: should we move this into `components/Buttons.tsx`?

src/components/Buttons.tsx
```diff
@@ -274,6 +274,7 @@ export const BigAsyncButton = forwardRef<HTMLButtonElement, AsyncButtonProps>(
     );
   },
 );
+
```
varortz: wait we should probably remove this line

src/lib/icons.tsx
```diff
@@ -216,6 +216,34 @@ export const IconSvgs = {
       />
     </svg>
   ),
+  IJPLogo: (
```
jinkang-0: this should be a webp, since it uses the image tag

src/components/NavBar/index.tsx
```diff
@@ -97,7 +96,7 @@ export default function NavBar() {
     <Styles.NavBarContainer>
       <Styles.NavBarSectionDiv>
         <Link href="/">
-          <Icon type="logo" />
+          <Icon type="IJPLogo" />
         </Link>
         {navlinks.map(NavLink => renderLink(NavLink))}
```
jinkang-0: let's put the definition of `renderLink` in here instead of outside the function.
also, remember to add the `key` prop to the outermost wrapper of the render function to avoid this warning:
image

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+import Icon from '@/components/Icon';
+import COLORS from '@/styles/colors';
+import { Flex } from '@/styles/containers';
+import { useProfile } from '@/utils/ProfileProvider';
+import { ProfileButton, SmallLinkButton } from '../Buttons';
+import * as Styles from './style';
+
+export default function NavBar() {
+  const profile = useProfile();
+  const AuthButtonView = useMemo(() => {
```
jinkang-0: nit: can we add a newline between the two to separate the hook types?

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (!profile) throw new Error('Profile must be defined.');
+    if (profile.profileReady && profile?.profileData)
+      return (
+        <ProfileButton href="/settings">
+          {profile.profileData?.first_name || 'Profile'}
+        </ProfileButton>
+      );
```
jinkang-0: this is mb, the check should instead be `auth.userId` - which means we need to `useAuth` here (note: and the corresponding error check)

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+import COLORS from '@/styles/colors';
+import { Flex } from '@/styles/containers';
+import { useProfile } from '@/utils/ProfileProvider';
+import { ProfileButton, SmallLinkButton } from '../Buttons';
+import * as Styles from './style';
+
+export default function NavBar() {
+  const profile = useProfile();
+  const AuthButtonView = useMemo(() => {
+    if (!profile) throw new Error('Profile must be defined.');
```
jinkang-0: let's place this error check right below where we first defined `profile`

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+          flex: 1,
+        }}
+      >
+        <Styles.NoUnderlineLink
+          href={link.path}
+          $color="white"
+          $isActive={link.active}
+        >
+          {link.name}
+        </Styles.NoUnderlineLink>
```
jinkang-0: unfortunately, the changing font weights will cause the layout of the navbar to shift as bolded text take up more space

to remedy this, we can either fix the width of each box, or try a more...
<details>
<summary>unorthodoxed approach</summary>

- make a duplicate of this text with the same font styles
- group the two in a box with position relative
- make one of the text absolute positioned
    - call the absolute positioned text "display text"
    - call the other one "layout text"
- make sure the two texts line up and share the same center
    - this can be achieved by giving the display text the following styles:
        - `top: 50%`
        - `left: 50%`
        - `transform: translate(-50%, -50%)`
    - and centering the layout text as normal
- make the layout text bolded, always
- give the display text the conditional bolding (i.e. with `$isActive`)
</details>

src/lib/icons.tsx
varortz: nice catch!

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </Flex>
+      <Styles.ActiveUnderline $isActive={link.active} />
+    </Styles.LinkContainer>
+  );
+
+  const currentPath = usePathname();
+  if (currentPath.includes('/onboarding')) {
```
jinkang-0: this looks like it'll include the roles page as well, but the navbar should still be present in the roles page
image

src/components/NavBar/index.tsx
```diff
@@ -1,20 +1,26 @@
...
+import IJPlogo from '~/public/images/ijp-logo.webp';
 import { ProfileButton, SmallLinkButton } from '../Buttons';
 import * as Styles from './style';
 
 export default function NavBar() {
   const profile = useProfile();
+  if (!profile) throw new Error('Profile must be defined.');
+
+  const auth = useAuth();
+  if (!auth) throw new Error('Auth Must be defined.');
```
varortz: nit: let's make "Must" all lowercase

src/components/NavBar/index.tsx
```diff
@@ -1,20 +1,26 @@
...
+
   const AuthButtonView = useMemo(() => {
```
varortz: I think the dependency array is missing `auth.userId`

public/images/ijp-logo.webp
jinkang-0, line 1: this looks a bit blurry, which could be due to small resolution - could you try exporting the png from figma with 4x enlargement before converting it into a webp?

src/components/Buttons.tsx
```diff
@@ -102,6 +102,19 @@ export const LinkButton = styled(Link)`
   text-decoration: none;
 `;
 
+export const SmallLinkButton = styled(Link)`
+  ${ButtonStyles}
+  text-decoration: none;
+  padding: 0.25rem 0.75rem;
+  font-weight: 400;
+  border-radius: 0.313rem; // 5px
```
jinkang-0: let's use px here for border radius and padding!

src/components/NavBar/style.ts
```diff
@@ -0,0 +1,72 @@
...
+`;
+
+export const AuthButtons = styled.div`
+  display: flex;
+  gap: 1rem;
+  align-self: center;
+  top: 0;
+  right: 0;
+`;
+export const NoUnderlineLink = styled(LinkColored)<{ $isActive: boolean }>`
```
jinkang-0: we can remove the `$isActive` prop since it's no longer used

jinkang-0: also nit: can we add a newline before this so we can separate the definition of `AuthButtons` from that of `NoUnderlineLink`?

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+        $direction="column"
+        $justify="center"
+        style={{
+          flex: 1,
+        }}
+      >
+        <Styles.NoUnderlineLink
+          href={link.path}
+          $color="white"
+          $isActive={link.active}
```
jinkang-0: we can remove the `$isActive` prop here since it's no longer used

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+import * as Styles from './style';
+
+export default function NavBar() {
+  const profile = useProfile();
+  if (!profile) throw new Error('Profile must be defined.');
+
+  const auth = useAuth();
+  if (!auth) throw new Error('Auth Must be defined.');
+
+  const AuthButtonView = useMemo(() => {
```
jinkang-0: let's use camel case for this since this is not technically a component constructor (but rather just TSX)

jinkang-0 (commented):
looks amazing kevin! just left some nits to polish up and it should be pretty much good to merge! thanks for hashing all the comments out!

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+          </Styles.DisplayText>
+          {link.name}
+        </Styles.NoUnderlineLink>
+      </Flex>
+      <Styles.ActiveUnderline $isActive={link.active} />
+    </Styles.LinkContainer>
+  );
+
+  const currentPath = usePathname();
+  if (currentPath.includes('/onboarding') && !currentPath.includes('/roles')) {
```
jinkang-0, line 100: hmm seems like the onboarding roles page still doesn't have a navbar
note: the URL for the onboarding roles page is `/onboarding/roles`

jinkang-0, line 91: nit: it might be better to do `... && !currentPath === '/onboarding/roles'` so this doesn't delete the navbar from new pages that include `/roles`

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+          Sign Up
+        </SmallLinkButton>
+      </>
+    );
+  }, [profile]);
+
+  function useActiveStatus(path: string): boolean {
+    const currentPath = usePathname();
+    return currentPath.includes(path);
+  }
```
jinkang-0: let's put the definition of this custom hook in `utils/hooks.ts`

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+      path: '/limited-case-assignments',
+      active: useActiveStatus('/limited-case-assignments'),
+    },
+    {
+      name: 'Language Support',
+      path: '/language-support',
+      active: useActiveStatus('/language-support'),
+    },
+  ];
+
```
jinkang-0: for static data like this, let's move it outside the function so it doesn't get re-computed when the navbar re-renders

note: every time a component re-renders, everything inside the component definition (except those wrapped in memo functions, for example) is re-run! so putting static data like this here would be pretty inefficient

varortz: We can move the type definition outside of the function, but we get this lint error when moving `navlinks`:
 
image

public/images/ijp-logo.webp
varortz: Here's a before and after of this change (is there an improvement?? I can't really tell):

**Before** 
image

**After (4x enlargement)**
image

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+          {profile.profileData?.first_name || 'Profile'}
+        </ProfileButton>
+      );
+
+    return (
+      <>
+        <SmallLinkButton
+          $primaryColor={COLORS.blueMid}
+          $secondaryColor={COLORS.blueDark}
+          href="/login"
```
ccatherinetan: do we want to use CONFIG here? (I believe they were added in the auth styles PR, which hasn't been merged yet, so we may have to add this to CONFIG again in this PR)

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+        >
+          Log In
+        </SmallLinkButton>
+        <SmallLinkButton
+          $primaryColor="white"
+          $secondaryColor={COLORS.blueDark}
+          href="/signup"
```
ccatherinetan: do we want to use CONFIG here? (I believe they were added in the auth styles PR, which hasn't been merged yet, so we may have to add this to CONFIG again in this PR)

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+        </SmallLinkButton>
+      </>
+    );
+  }, [profile, auth.userId]);
+
+  const navlinks: NavLink[] = [
+    { name: 'Cases', path: '/cases', active: useActiveStatus('/cases') },
+    {
+      name: 'Limited Case Assignments',
+      path: '/limited-case-assignments',
```
ccatherinetan: do we want to use CONFIG here? (and for /language-support, /cases)

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  return (
+    <Styles.NavBarContainer>
+      <Styles.NavBarSectionDiv>
+        <Link href="/">
+          <Image
+            alt="background"
+            src={IJPlogo.src}
+            placeholder="blur"
+            blurDataURL={IJPlogo.src}
```
ccatherinetan: Is it necessary to have the blurDataURL, if it is just the same as the src?

src/components/NavBar/index.tsx
jinkang-0: ah, i think it's better not to use hooks more than once if we can help it.
here, it may be better to just get the pathname using `usePathname` and here, instead of the hook, make an anonymous arrow function like this:
```ts
(pathname: string) => pathname === '/language-support'
```
then call that function in the rendering function, passing in the current pathname

src/components/NavBar/index.tsx
varortz: yeah this can probably get addressed while rebasing! It should be on main now.

public/images/ijp-logo.webp
jinkang-0: this is how it looks for me
image

varortz: Ah, I can add the one exported with 4x enlargement, and then maybe you can check if it looks better on your end? For some reason, I'm not seeing a difference on my end.

src/components/NavBar/index.tsx
varortz: Is there an issue with calling the hook multiple times? I feel like using the hook may be better for readability.

jinkang-0: Primary issue is redundancy. I don't know how `usePathname` impacts performance, but it feels wasteful to call it multiple times when we can just call it once and store it. Further, doing it the other way allows us to store the nav links data outside of the component function, so we can avoid re-defining it each time this component re-renders.

This is ultimately a nit, and I agree that this may only offer slight performance gains. However, I do think it can be better in terms of performance and I don't think it will sacrifice much in terms of readability (we just turn the status from a variable into a function).

varortz: Yeah, I agree that it's best to not call `usePathname` multiple times if we can avoid it. Thoughts on this approach instead of creating a function:


``` 
  const currentPath = usePathname();

...
...
...

  const navlinks: NavLink[] = [
    { name: 'Cases', path: CONFIG.cases, active: currentPath === CONFIG.cases },
    {
      name: 'Limited Case Assignments',
      path: CONFIG.lca,
      active: currentPath === CONFIG.lca,
    },
    {
      name: 'Language Support',
      path: CONFIG.languageSupport,
      active: currentPath === CONFIG.languageSupport,
    },
  ];;```

jinkang-0: I noticed a pattern of the `path` being the same as the comparison. Could we just add a check like `currentPath === link.path` in the rendering function? Then we can move the nav links definition outside of the component function

src/components/NavBar/index.tsx
varortz: not sure, it seems like `blurDataURL` acts as a placeholder for the actual image (`src`). Here is a screenshot of Next.js' documentation:

image

Should we consider removing the `blurDataURL` placeholder if we're using the same image as the placeholder and it can impact performance since these source images are likely not 10px lol??

varortz: Wait, based on Next.js' documentation it seems like `blurDataURL` would be automatically populated in this case:

image

We should be fine with removing `blurDataURL` entirely.

Here is a link to a quick video on it as well: https://www.youtube.com/watch?v=Bz3No1RFXWY

jinkang-0: For it to be automatically populated, we have to pass in `IJPLogo` entirely to `src`. Not just the source path.

varortz: ah ok so `src={IJPlogo}` and we remove `blurDataURL`? Should we move forward with this for all other images then?

jinkang-0: If it works, I'd be down! Let's experiment with it a little bit first though, I'm also not entirely sure

varortz: ah nvm let's keep it as is for now. Here is a video of the effect of applying this change (change applied to the IJP logo on the navbar): 

https://github.com/calblueprint/immigration-justice-project/assets/84427964/4ea31d35-a35c-448a-9a14-3380185467cc

src/components/NavBar/index.tsx
varortz: Wait I don't know how I didn't see that before!!

public/images/ijp-logo.webp
varortz: @jinkang fix(?) should be pushed, can you check if it looks fine on your end?

jinkang-0: pushed a commit to fix this - i think the issue was that Next.js's `Image` tag sets the image resolution directly and doesn't downscale; changing it to be a style of width and height fixed it for me

jinkang-0 (approved):
lgtm!

varortz (approved):
NAVBAR LOOKS AMAZING!!! Thank you Kevin for working on this throughout semester, you killed it :)

PR #48, opened 2023-11-29
[feat] adds case details to the CaseDetails component

varortz:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

## :tanabata_tree: Description
Adds case details for the selected case listing card to the cases page.
### :palm_tree: What's new in this PR
[//]: # "REQUIRED - Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."
- Centers line dividing the `CardColumn` & the` InterestForm` + `Case Details`
- Adds case fields to the case details 
- Implements helper for parsing and formatting some of the case fields
- Fixes width of radio button borders

### :evergreen_tree: Screenshots
[//]: # "REQUIRED for frontend changes, otherwise optional but strongly recommended. Add screenshots of expected behavior - GIFs if you're feeling fancy! If you are making changes here, please CC: @davidqing6432 at the bottom."

image

## :deciduous_tree: How to review
[//]: # 'REQUIRED - Describe the order in which to review files and what to expect when testing locally. Is there anything specifically you want feedback on? Should this be reviewed commit by commit, or all at once? What are some user flows to test? What are some edge cases to look out for?'
- Go to the cases page and compare it to the figma designs

## :seedling: Next steps 
[//]: # "Optional - What's NOT in this PR, doesn't work yet, and/or still needs to be done. Note any temporary fixes in this PR that should be cleaned up later."

### :information_source: Online sources
[//]: # 'Optional - copy links to any tutorials or documentation that was useful to you when working on this PR'

### :potted_plant: Related PRs
[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc; if this is a refactor, feel free to add PRs that previously modified this code"

[//]: # 'This tags the project leader as a default. Feel free to change, or add on anyone who you should be in on the conversation.'
CC: @pragyakallanagoudar

PR #68, opened 2024-03-10
[feat] homepage implementation

kevin3656:

Implemented the Homepage

- Added styles.tsx for homepage
- implemented homepage
- added color goldMid to COLORS
- added images folder and stock photo images

Screen Shot 2024-03-10 at 2 03 22 PM
Screen Shot 2024-03-10 at 2 03 41 PM
Screen Shot 2024-03-10 at 2 03 38 PM

- should appear on localhost:3000 when running npm
- would like feedback on overall structure of Homepage

## :seedling: Next steps 

- finalizing contents of homepage(statistics, descriptions in "How You Can Help" and Icons
- adding functionality to buttons
- drop shadow on statistics section

CC: @varortz


src/styles/colors.ts
```diff
@@ -18,6 +18,8 @@ const COLORS = {
   redDark: '#B53839',
   redLight: '#F3B1B2',
   redLighter: '#FFE9E9',
+  // Gold
```
varortz: let's remove this this comment!

src/app/page.tsx
```diff
@@ -1,9 +1,137 @@
...
+          <StatLabel>Attorneys</StatLabel>
+        </StatContainer>
+        <StatContainer>
+          <StatNumber>100+</StatNumber>
+          <StatLabel>Translation Support</StatLabel>
+        </StatContainer>
+        <StatContainer>
+          <StatNumber>100+</StatNumber>
+          <StatLabel>Succesful Outcomes</StatLabel>
+        </StatContainer>
```
varortz: Could we simplify this code so that we only use the `StatContainer`, `StateNumber`, and `StatLabel` components once, and we map the stat number and stat label values into these components? For an example, you can refer to[ this](https://github.com/calblueprint/immigration-justice-project/pull/48/files#diff-fa3adc90665e593292f8a500fadfa2f0eefc08f6849f621353240a413d190657) PR where a something similar is done with the `caseFields` & `renderField`.

src/app/page.tsx
```diff
@@ -1,9 +1,137 @@
...
+              src="path_to_case_assistance_icon"
+              alt="An Icon Will Be here"
+            />
+            <ServiceTitle>Translation Assignment</ServiceTitle>
+            <ServiceDescription>
+              Interpreters and translators can conduct volunteer translation and
+              live interpretation. We have frequent need for individuals who
+              speak Creole, Portuguese, Spanish, French, Arabic, and Russian.
+            </ServiceDescription>
+          </ServiceContainer>
```
varortz: Similarly, here we have a pattern of `ServiceTitle`, `ServiceDescription`, `ServiceContainer`, and `ServiceIcon` components. Could we also simplify this as mentioned in [this](https://github.com/calblueprint/immigration-justice-project/pull/68/files#r1521953465) comment above.

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+`;
+
+// Label for each stat
+export const StatLabel = styled.div`
+  font-size: 22px;
+  color: ${COLORS.greyDark};
+`;
+export const StatContainer = styled.div`
+  text-align: center;
+  /* padding-left: 374px;
```
varortz: could we remove these commented out lines of code?

src/styles/text.tsx
```diff
@@ -42,6 +42,14 @@ export const H4 = styled.h4<{ $color?: string }>`
   margin: 0;
 `;
 
+export const H5 = styled.h5<{ $color?: string }>`
```
varortz: For our headers, we try to have `H1` have the largest font size, and `H2` would have the next largest font size, and so on. Using this convention,  `H5` would need to have a smaller font size, so we should probably rename this. Additionally, since it seems like this large font size is only being used in the home page so far, we can remove it from this file and move it to the `styles.tsx` file corresponding to the home page (maybe rename it it to `Title`).

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+`;
+
+// Service title
+export const ServiceTitle = styled.h2`
+  font-size: 20px;
+  color: ${COLORS.blueMid};
+`;
+
+// Service description
+export const ServiceDescription = styled.p`
```
varortz: in `styles/text.tsx`, `H4` has font size 16. Let's try to use the global styles instead!

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+`;
+
+// Service icon
+export const ServiceIcon = styled.img`
+  height: 50px;
+  margin-bottom: 10px;
+`;
+
+// Service title
+export const ServiceTitle = styled.h2`
```
varortz: Since this is very similar in font size to our `H3` in `styles/text.stx` could we go ahead and use the global styling instead?

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+  height: 557px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 10px;
+  padding-top: 70px;
+`;
+
+// Title and subtitle
+export const Title = styled.h1`
```
varortz: This is a one `px` difference from the `H2` header defined in `styles/text.tsx`. Could we go ahead and replace this with `H2` on the home page, and remove `Title` from here?

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+  ${openSans.style}
+  font-size: 35px;
+  font-weight: 500px;
+  color: ${COLORS.blueMid};
+`;
+
+export const Subtitle = styled.p`
```
varortz: we have `H4` defined in `styles/text.tsx` with font size of `16px` (`1rem`). Let's go ahead and remove this and use `H4` on the home page instead.

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+`;
+
+export const StatisticsSection = styled.div`
+  width: 100%;
+  height: 146;
+  background-color: white;
+  display: flex;
+  justify-content: space-around;
+  align-items: center;
+  box-shadow: 0px 8px 20px rgba(0, 0, 0, 0.2);
```
varortz: `box-shadow` seems to not be working since `StatisticsSection` and the `HowYouCanHelpContainer` are sibling elements. Could we try to find a fix in this PR?

src/app/page.tsx
```diff
@@ -1,9 +1,137 @@
-import styles from './page.module.css';
```
jinkang-0, line 1: can we delete the file also

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+  padding-top: 33px;
+  padding-bottom: 35px;
+`;
+export const StatNumber = styled.div`
+  font-size: 35px;
```
jinkang-0: For this and all other font sizes, please use `rem`.
More ideally, don't define it at all, and reuse text components defined in `styles/text.tsx`.

src/app/page.tsx
```diff
@@ -1,9 +1,137 @@
...
+            <ServiceDescription>
+              Interpreters and translators can conduct volunteer translation and
+              live interpretation. We have frequent need for individuals who
+              speak Creole, Portuguese, Spanish, French, Arabic, and Russian.
+            </ServiceDescription>
+          </ServiceContainer>
+        </ServicesDiv>
+      </HowYouCanHelpContainer>
+      <MissionValuesContainer>
+        <img src={image.src} alt="Smaller Stock" width={521} height={338} />
```
jinkang-0: I would recommend using the [Image component](https://nextjs.org/docs/pages/api-reference/components/image) provided by Next.js to help optimize it

jinkang-0: Ideally, we shouldn't fix arbitrary pixel values to width and height, but rather use units that help scale it relative to the viewport. In this case, I would recommend using a `grid` layout and specifying a reasonable fractional unit for the left column.

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+  padding-top: 200px;
+  padding-right: 440px;
+  padding-left: 84px;
+  gap: 30px;
+`;
+
+export const ButtonDiv = styled.div`
+  width: 244px;
+  height: 40px;
+`;
```
jinkang-0: It's good practice to not fix `width` and `height` as screen sizes differ drastically. Rather, a better solution would be to use fractional units with pixel units as a boundary. For example, `width: clamp(224px, 25%, 500px)`.

src/styles/text.tsx
```diff
@@ -22,11 +22,11 @@ export const H2 = styled.h2<{ $color?: string }>`
   margin: 0;
 `;
 
-export const H3 = styled.h3<{ $color?: string }>`
+export const H3 = styled.h3<{ $color?: string; $fontWeight?: number | string }>`
   ${openSans.style}
   font-size: 1.5rem;
-  font-weight: 600;
   color: ${props => props.$color || COLORS.greyDarker};
+  font-weight: ${props => props.$fontWeight || '600'};
```
jinkang-0: I believe we decided to use the practice of destructuring. Thus, here we would do `font-weight: ${({ $fontWeight }) => $fontWeight || '600'}` instead. The `color` prop is an artifact from a time long past, it would be great if you can update that to use destructuring as well 👍

src/styles/text.tsx
jinkang-0, line 1: @kyrenetam can you redefine the header tags for the new design? It would help a lot with staying consistent in devving. For example, you can see what David decided on last semester in the Figma:

image

src/app/styles.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export const TitleSection = styled.div`
+  width: 100%;
+  height: 100vh; // Use 100vh to make the div fill the entire viewport height
+  background: linear-gradient(
+      222deg,
+      rgba(40, 120, 194, 0.5) 14.2%,
+      rgba(0, 50, 96, 0.5) 76.27%
+    ),
+    url(${image.src}) no-repeat center / cover;
```
jinkang-0: Instead of a background image, I would strongly recommend simply putting the Image tag in the HTML and overlaying the text on top of it using `position: absolute` and setting the `z-index` to be some positive number. Background images do not take advantage of Next.js' image optimization and the lack of a background image could look really nasty on devices with slower internet.

jinkang-0 (commented):
looks good so far Kevin! I didn't have a chance to look at it running locally, but I think there were also some changes you haven't pushed, so I'll delay that review for another time.

jinkang-0:
side note: we generally follow the PR naming convention of `[tag] lowercased description of pr`, would be great if we can stay consistent with your most recent PRs as well!

src/app/styles.tsx
varortz: we can instead use the `H2` header form `styles/text.tsx`

src/app/page.tsx
```diff
@@ -1,9 +1,137 @@
...
+        <H5>IMMIGRATION JUSTICE PROJECT</H5>
+        <H3 $color="white" $fontWeight="400">
+          The Immigration Justice Project (IJP) seeks to promote due process and
+          access to justice through the provision of high quality legal services
+          on behalf of indigent immigrants and asylum seekers.
+        </H3>
+        <ButtonDiv>
+          <Button
+            $primaryColor={COLORS.goldMid}
+            $secondaryColor={COLORS.blueDarker}
```
varortz: @kyrenetam on hover for gold buttons, do we want the color of the button to change to a darker gold perhaps?

src/app/page.tsx
```diff
@@ -1,9 +1,166 @@
...
+  ServiceContainer,
+  ServiceIcon,
+  ServicesDiv,
+  MissionValuesContainer,
+  MissionStatement,
+  TextContainer,
+  ImageBackground,
+  ImageContainer,
+  ImageWrapper,
+} from './styles';
```
jinkang-0: I would suggest to consolidate this into `import * as Styles from './styles'` and use the styled components like `<Styles.StatLabel>...</Styles.StatLabel>`

This is up for discussion, but I've been enjoying this paradigm as it reduces the number of imports and also reduces the friction of adding/removing imports from the styles file later on

src/app/page.tsx
```diff
@@ -1,9 +1,166 @@
...
+  { number: '100+', label: 'Successful Outcomes' },
+];
+
+const renderStat = (stat: Stat) => (
+  <StatContainer key={stat.label}>
+    <H2 $color={COLORS.blueDark}>{stat.number}</H2>
+    <StatLabel>{stat.label}</StatLabel>
+  </StatContainer>
+);
+type Service = {
```
jinkang-0, line 30: nit: can we add a newline to separate the two?

src/react-app-env.d.ts
jinkang-0, line 1: minor nit: should we name this `image-declaration.d.ts` to make the purpose of this clearer?

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
+import styled from 'styled-components';
+import COLORS from '@/styles/colors';
+import { openSans } from '@/styles/fonts';
+
+export const PageContainer = styled.div`
+  width: 100%;
+  min-height: 100vh;
```
jinkang-0: nit: let's use `100svh` so it doesn't get covered by the bottom tab of the mobile browser

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+`;
+
+export const TitleSection = styled.div`
+  position: relative;
+  width: 100%;
+  min-height: 100vh; // Use 100vh to make the div fill the entire viewport height
```
jinkang-0: nit: let's use `height: 100svh` instead of `min-height`

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  display: flex;
+  flex-direction: column;
+  overflow-y: hidden;
```
jinkang-0: q: why do we need an `overflow-y: hidden` here? If it's for mobile view, it may be better to address the size of content inside the div rather than cutting it off. If that's not possible, it seems better to me to just let it overflow instead of cutting off the content entirely.

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+export const ButtonDiv = styled.div`
+  margin-top: 25px;
+  width: clamp(224px, 25%, 500px);
+  height: clamp(40px, 5vh, 60px);
+`;
+
+export const StatisticsSection = styled.div`
+  width: 100%;
+  height: 146;
```
jinkang-0: let's specify a unit here

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  display: flex;
+  flex-direction: column;
+  overflow-y: hidden;
+`;
+
+export const ButtonDiv = styled.div`
+  margin-top: 25px;
+  width: clamp(224px, 25%, 500px);
+  height: clamp(40px, 5vh, 60px);
+`;
```
jinkang-0: Upon inspection, it seems that the button has a font size of `16px` instead of the `24px` specified on Figma. Maybe we can add `font-size: 1.5rem !important` here to force it to `24px`

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+// Service container
+export const ServicesDiv = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  margin-top: 60px;
+  gap: 90px;
+  padding-left: 190px;
+  padding-right: 190px;
+`;
```
jinkang-0: Can you add a `clamp` here to allow the padding to decrease dynamically as the screen sizes changes? It may also help to add a `width: 100%` here so it doesn't overflow and cause the rest of the page to be offset.

Protip: you can shorten this to `padding: 0 190px`

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  grid-template-columns: 1fr 2fr; // Example adjustment
+  grid-template-rows: auto;
+  gap: 20px; // Adjust space between grid items
+  background-color: ${COLORS.blueLighter};
+  width: 100%;
+  min-height: 488px;
+  align-items: center;
+  justify-content: space-around;
+  gap: 50px;
+`;
```
jinkang-0: Let's add a `padding: 73px 128px` to better match what's on Figma. Of course, we should also specify some clamping to adjust for the mobile view - I'll leave the specifics up to your discretion.

jinkang-0: Also, it seems you specified `gap` twice. Let's remove the extra, and set it to `gap: 40px`, which is what's specified on the Figma.

src/app/page.tsx
```diff
@@ -1,9 +1,166 @@
...
+  },
+];
+
+const renderService = (service: Service) => (
+  <ServiceContainer key={service.title}>
+    <ServiceIcon src={service.iconSrc} alt={service.title} />
+    <H3 $color={COLORS.blueMid}>{service.title}</H3>
+    <H4 $color={COLORS.greyDark} $fontWeight="400">
+      {service.description}
+    </H4>
```
jinkang-0: I think it may be better to use `<H4 ... dangerouslySetInnerHTML={{ __html: service.description }} />` and change the `description` field to be a JSX element (i.e. `<><b>Interpreters and translators</b> can ...</>`)

This will allow us to match the bolded words in Figma and allow us to use the generalization method you have implemented here.

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+export const HowYouCanHelpContainer = styled.div`
+  background-color: ${COLORS.background};
+  width: 100%;
+  min-height: 557px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 10px;
+  padding-top: 70px;
+`;
```
jinkang-0: No need to specify height, it should be automatically scaled depending on the size of the content. Arbitrary specifications may harm usability if the user (for whatever reason) sets the font size to be smaller than usual.

Also, `padding-top` should be `58px`, as according to Figma. Background color should be specified in the layout body (cf: [changes to layout in this PR](https://github.com/calblueprint/immigration-justice-project/pull/73/files#diff-29d0a55f5ca1a9951042587f8b891bb4cd8b204f5bc855dc2468ed44cde2f5b3))

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  min-height: 488px;
+  align-items: center;
+  justify-content: space-around;
+  gap: 50px;
+`;
+export const MissionStatement = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  max-width: 500px;
```
jinkang-0: No need to specify width here, it should be handled by the parent.

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  gap: 10px;
+  padding-left: 50px;
```
jinkang-0: Padding should be handled by the parent.

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  top: 0;
+  left: 0;
+  max-width: 900px;
+  padding-top: 200px;
+  padding-left: 84px;
+  padding-bottom: 150px;
+  z-index: 1;
+  padding-right: 32px;
+  gap: 50px;
+  flex-wrap: wrap;
```
jinkang-0: Let's use `gap: 30px` to match Figma.
We don't need `padding-right` here, so we can get rid of it.
For `max-width` and the rest of the paddings, let's add some clamping with reasonable values to address for mobile view.

If you need to play around with how it looks like on mobile, open up the web tool (F12 or right click and inspect element) and either resize the toolbar so it's close to mobile width (~400px) or click on the icon that looks like a laptop and phone to adjust the device size.

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  left: 0;
+  height: 100%;
+  width: 100%;
+  background-color: rgba(40, 120, 194, 0.5);
+  z-index: 1;
+`;
+export const ImageContainer = styled.div`
+  position: absolute;
+  width: 100%;
+  height: 100vh;
```
jinkang-0: nit: let's use `100svh`

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+  top: 0;
+  left: 0;
+`;
+export const Title = styled.h5<{ $color?: string }>`
+  ${openSans.style}
+  font-size: 5.31rem;
+  font-weight: 600;
+  color: white;
+  margin: 0;
+`;
```
jinkang-0: let's add `line-height: 5.5rem` and move this to `@/styles/text.tsx`

src/app/styles.tsx
jinkang-0, line 1: nit: since we're not defining any JSX/TSX here, let's rename the file to `styles.ts`.

src/app/styles.tsx
```diff
@@ -0,0 +1,145 @@
...
+export const ImageWrapper = styled.div`
+  grid-area: 1 / 2;
```
jinkang-0: you shouldn't need to specify the `grid-area` since it'll automatically determine that based on the order that the elements are defined, so we can safely remove this

src/app/styles.tsx
varortz: If we remove the `overflow-y`, the content will overlap with other sections of the home page which can make it look pretty messy. However, even now it behaves a little weird because everything within `TextContainer` will simply just hide behind the `StatisticsSection` with `overflow-y: hidden`.

jinkang-0: Yes, it's better not to arbitrarily overflow it, and instead account for the weird format itself - that's the root of the issue.

src/app/styles.ts
```diff
@@ -0,0 +1,145 @@
...
+// Label for each stat
+export const StatLabel = styled.div`
+  font-size: 1.375rem;
+  color: ${COLORS.greyDark};
+`;
+export const StatContainer = styled.div`
+  text-align: center;
+`;
+export const HowYouCanHelpContainer = styled.div`
+  width: 100vw;
```
jinkang-0: please use `100%` so it doesn't overflow

`100vw` will make this fill the entire viewport width, but if there's a vertical scrollbar with definitive width (i.e. windows on chrome), then the `100vw` container will overlap with the scrollbar, causing a horizontal overflow.

public/HomePageImage.webp
jinkang-0, line 1: let's put this in `public/images/` just for sake of organization

src/app/styles.ts
```diff
@@ -0,0 +1,145 @@
...
+`;
+
+export const TextContainer = styled.div`
+  max-width: 900px;
+  max-height: 395px;
+  padding-top: clamp(5px, 25vh, 200px);
+  padding-left: clamp(40px, 10%, 84px);
+  z-index: 1;
+  gap: 50px;
+`;
```
jinkang-0: ah mb, it should be `gap: 30px`
however, for it to apply, you need to add `display: flex;` 
and for sake of orientation, make sure to also add `flex-direction: column;`

src/app/page.tsx
```diff
@@ -1,9 +1,153 @@
...
+          <Image
+            alt="background"
+            src="/HomePageImage.webp"
+            placeholder="blur"
+            blurDataURL="/HomePageImage.webp"
+            quality={100}
+            layout="fill"
+            objectFit="cover"
+            fill
+          />
```
jinkang-0: for this and the `Image` in the mission statement, use a statically imported image

i.e.
```
import homepageImage from '../../public/HomePageImage.webp';

...

<Image
  ...
  src={homepageImage.src}
  ...
/>
```

src/app/styles.ts
```diff
@@ -0,0 +1,145 @@
...
+
+// Service icon
+export const ServiceIcon = styled.img`
+  height: 50px;
+  margin-bottom: 10px;
+`;
+
+export const MissionValuesContainer = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 2fr; // Example adjustment
```
jinkang-0: the design is more like `45fr 55fr`, let's adjust that

jinkang-0 (commented):
added some comments to address issues of images and some spacing
make sure to go back and address some older comments as well, there were some that were not resolved fully

good progress so far though!
everything is almost identical to the design 👍

src/react-app-env.d.ts
jinkang-0, line 1: forwarding past comment: 
can we name this file `image-declaration.d.ts` just to be more clear about what this is for?
the name won't matter for Typescript, so it's better to name it something we understand

src/app/page.tsx
```diff
@@ -1,9 +1,154 @@
...
+  number: string;
+  label: string;
+};
+
+const stats: Stat[] = [
+  { number: '181+', label: 'Matched Cases' },
+  { number: '181+', label: 'Attorneys' },
+  { number: '65+', label: 'Interpreters & Translators' },
+  { number: '46+', label: 'Successful Outcomes' },
+];
```
jinkang-0: @varortz @kyrenetam what's the plan for the stats?
are we going for 180+, 180+, 65+, 45+ or 181, 181, 66, 46?

src/app/page.tsx
```diff
@@ -1,9 +1,154 @@
...
+          <Image
+            alt="background"
+            src={homepageImage.src}
+            placeholder="blur"
+            blurDataURL={homepageImage.src}
+            quality={100}
+            layout="fill"
+            objectFit="cover"
+            fill
+          />
```
jinkang-0: let's remove the `objectFit="cover"` here as it's a legacy prop
instead, we can add `style={{ objectFit: 'cover' }}`

src/app/page.tsx
```diff
@@ -1,9 +1,154 @@
...
+        <Styles.ImageWrapper>
+          <Image
+            alt="Lawyer Image"
+            src={homepageImage.src}
+            placeholder="blur"
+            blurDataURL={homepageImage.src}
+            quality={100}
+            layout="fill"
+            objectFit="contain"
+          />
```
jinkang-0: similarly for this image: `objectFit="contain"` -> `style={{ objectFit: 'contain' }}`

src/app/page.tsx
varortz: plus signs! We're going with this approach given that the stats we have are based on data from their airtable (2020 onwards), and they mentioned they had more data somewhere else so we should maybe account for that missing data with the `+` signs

jinkang-0: should we use 180+, 180+, 65+, and 45+?
for some reason a non-multiple of 5 looks weird

varortz: I don't really have a strong opinion on this. I see your point, but I also feel like it can make the stats feel more legit lol. @kyrenetam @kevin3656 what do y'all think? I'm ok with either approach.

varortz: Oh wait (I forgot to bring this up during last worksesh), Ambreen also asked if we could add functionality so that admin can update these stats through Retool. Definitely something we should implement if we have time!

jinkang-0 (approved):
lgtm

varortz (approved):
AMAZING WORK KEVIN!! we have a home page now :))

PR #82, opened 2024-05-01
[feat] implement conditional routing from Volunteer Now

ccatherinetan:

* implements conditional routing from Volunteer Now button on home page. 
* implements conditional routing in login
    * (an already logged in user will be redirected to settings)
    * routing to listing page / onboarding after logging in 

routing logic 
* if not signed in → Log in 
* if not onboarded → Onboarding
* if Attorney (or Attorney + Interpreter) → Cases
* If Legal Fellow (or Legal Fellow + Interpreter) → LCA page
* else (if Interpreter only) → Language Support 

check `app/page`

## :seedling: Next steps 

CC: @varortz

### DIFF
--- a/src/app/(auth)/login/page.tsx
+++ b/src/app/(auth)/login/page.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useContext, useState } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { useRouter } from 'next/navigation';
 import isEmail from 'validator/lib/isEmail';
 import { H4Centered, SpacerDiv } from '@/app/(auth)/styles';
@@ -22,6 +22,18 @@ export default function Login() {
   const [passwordError, setPasswordError] = useState('');
   const { push } = useRouter();
   const validEmail = (e: string) => e !== '' && isEmail(e);
+
+  useEffect(() => {
+    if (auth && auth.userId) {
+      if (profile && profile.profileData) {
+        if (profile.roles.map(r => r.role).includes('ATTORNEY')) push('/cases');
+        else if (profile.roles.map(r => r.role).includes('LEGAL_FELLOW'))
+          push('/limited-case-assignments');
+        else push('/language-support');
+      }
+      push(CONFIG.onboardingHome);
+    }
+  }, [auth, profile, push]);
 
   const handleSignIn = async () => {
     if (!auth) {
--- a/src/app/(auth)/login/page.tsx
+++ b/src/app/(auth)/login/page.tsx
@@ -24,8 +24,7 @@ export default function Login() {
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
   useEffect(() => {
-    if (!auth || !auth.userId) return;
-    push(CONFIG.settings);
+    if (auth && auth.userId) push(CONFIG.settings);
   }, [auth, profile, push]);
 
   const handleSignIn = async () => {
@@ -50,7 +49,7 @@ export default function Login() {
     } else {
       profile?.loadProfile();
       setErrorMessage('');
-      // conditional routing after log in
+      // conditional routing after logging in
       if (!profile?.profileData) push(CONFIG.onboardingHome);
       else if (profile.roles.map(r => r.role).includes('ATTORNEY'))
         push(CONFIG.cases);
--- a/src/app/(auth)/login/page.tsx
+++ b/src/app/(auth)/login/page.tsx
@@ -24,15 +24,13 @@ export default function Login() {
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
   useEffect(() => {
-    if (auth && auth.userId) {
-      if (profile && profile.profileData) {
-        if (profile.roles.map(r => r.role).includes('ATTORNEY')) push('/cases');
-        else if (profile.roles.map(r => r.role).includes('LEGAL_FELLOW'))
-          push('/limited-case-assignments');
-        else push('/language-support');
-      }
-      push(CONFIG.onboardingHome);
-    }
+    if (!auth || !auth.userId) return;
+    if (profile && profile.profileData) {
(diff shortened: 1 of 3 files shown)


src/app/(auth)/login/page.tsx
```diff
@@ -22,6 +23,18 @@ export default function Login() {
...
+    if (auth && auth.userId) {
+      if (profile && profile.profileData) {
+        if (profile.roles.map(r => r.role).includes('ATTORNEY')) push('/cases');
+        else if (profile.roles.map(r => r.role).includes('LEGAL_FELLOW'))
+          push('/limited-case-assignments');
+        else push('/language-support');
+      }
+      push(CONFIG.onboardingHome);
+    }
+  }, [auth, profile, push]);
```
jinkang-0: nit on code readability: to reduce nesting, since nothing happens in the case that `auth && auth.userId` is false, we can instead add the reverse check (`!(auth || auth.userId)`) and return if that's true

edge case: let's add a else statement for `push(CONFIG.onboardingHome)`, since `push` itself doesn't return the control of the function (i.e. it could push to `language-support`, then `CONFIG.onboardingHome`).

src/app/(auth)/login/page.tsx
```diff
@@ -44,7 +57,13 @@ export default function Login() {
     } else {
       profile?.loadProfile();
       setErrorMessage('');
-      push('/cases');
+      // conditional routing after log in
+      if (!profile?.profileData) push(CONFIG.onboardingHome);
+      else if (profile.roles.map(r => r.role).includes('ATTORNEY'))
+        push('/cases');
```
jinkang-0: lowkey could be good to add `CONFIG` values for these as well, just so it's easier to refactor in case the NPO ever decides to change the page name

src/app/page.tsx
```diff
@@ -67,6 +71,20 @@ const renderService = (service: Service) => (
...
 
 export default function Home() {
+  const auth = useAuth();
+  const profile = useProfile();
+
+  const volunteerLink = useMemo(() => {
+    if (!auth) throw new Error('Auth must be defined.');
+    if (!auth.userId) return '/login';
+    if (!profile) throw new Error('Profile must be defined.');
+    if (!profile.profileData) return CONFIG.onboardingHome;
```
varortz: can we update this check to `profile.profileReady && !profile.profileData`? Otherwise, it always routes an already onboarded use to the onboarding flow instead of a listing page.

src/app/(auth)/login/page.tsx
```diff
@@ -22,6 +23,18 @@ export default function Login() {
...
 
+  useEffect(() => {
+    if (auth && auth.userId) {
+      if (profile && profile.profileData) {
+        if (profile.roles.map(r => r.role).includes('ATTORNEY')) push('/cases');
+        else if (profile.roles.map(r => r.role).includes('LEGAL_FELLOW'))
+          push('/limited-case-assignments');
+        else push('/language-support');
+      }
+      push(CONFIG.onboardingHome);
```
varortz: After login, I'm noticing it always redirects already onboarded users to the onboarding flow (and then settings page after realizing we're onboarded).

https://github.com/calblueprint/immigration-justice-project/assets/84427964/9ddd1595-d724-406d-a348-5f09f3a6bc06


If we wrap this line in an `else if` to check that`profile?.profileReady && !profile.profileData`, the bug seems to be fixed:




`else if (profile?.profileReady && !profile.profileData) 
       push(CONFIG.onboardingHome);
       `

Let's go ahead and update this so that it has the expected behavior!

src/app/(auth)/login/page.tsx
ccatherinetan: sounds good! i add cases, lca and language support routes to CONFIG. Did we want to add routes for all other pages too (e.g. login, signup)? I guess those are pretty standard, but just for consistency's sake? 
If so, I could add the login route since i'm using in this PR

src/app/(auth)/login/page.tsx
ccatherinetan: should be fixed now!

src/app/page.tsx
ccatherinetan: gotcha, changed! 
the behavior looks right now, but it's still very slow for me. i.e. when I click volunteer now with a completed profile, I'm routed to login first, and then to a listing page after a second. Do we hav any ideas for how to fix this? not sure if it's an issue with my conditional logic or with useMemo?

src/app/(auth)/login/page.tsx
jinkang-0: ye feel free to add them!

src/app/page.tsx
jinkang-0: could try changing the user id check to `profile.profileReady && !auth.userId`
and return `''` if `!profile.profileReady`

ccatherinetan: gotcha i've changed the implementation to this suggestion!

src/app/(auth)/login/page.tsx
```diff
@@ -44,7 +49,13 @@ export default function Login() {
     } else {
       profile?.loadProfile();
       setErrorMessage('');
-      push('/cases');
+      // conditional routing after logging in
+      if (!profile?.profileData) push(CONFIG.onboardingHome);
```
jinkang-0: should we add a check for `profile.profileReady` here as well?

src/app/(auth)/login/page.tsx
```diff
@@ -22,6 +23,10 @@ export default function Login() {
   const { push } = useRouter();
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
+  useEffect(() => {
+    if (auth && auth.userId) push(CONFIG.settings);
```
jinkang-0: nit: if `auth` is not defined, we should throw an error instead

jinkang-0 (approved):
lgtm!

varortz (approved):

PR #83, opened 2024-05-14
[fix] fixes routing bugs

varortz:

Fixes routing issue with the limited case assignments page that was causing the a `404 Page Not Found` error.

- fixes `CONFIG.lca` path 
- updates the "Go to Profile change any information" text to match the designs
- fixes bug where an already onboarded user is routed to onboarding then settings on a fresh load after login

**"Go to Profile change any information" text:**
image

- `git fetch` on `main` & checkout `alvaro/ijp-138-fix-routing-bugs`! -> then `npm run dev`
 - Log in and update your user role to include the either `Legal Fellow` or `Legal Fellow and Interpreter`
 - If your role already includes "Legal Fellow", you should have been redirected to the LCA page after logging in during the previous step. Otherwise, head to the settings page and sign out (You should be redirected to the Log In page).
 - Log in, and make sure that you are routed to the LCA page! This uses `CONFIG.lca`.
 
**ALSO PLEASE PLEASE TEST FOR OTHER ROUTING ISSUES IF YOU CAN, SO WE CAN ADDRESS THEM IN THIS PR** 🫶🏼

- For the Profile Match text, make sure it is consistent with the designs (Noticed we use "update" instead of "change", currently asking Kyrene about this + will check if this change was documented anywhere)
 
## :seedling: Next steps 

  - Potentially look into another approach for conditional routing (currently using a `useEffect` to achieve this)??

- CONFIG.lca bug introduced in https://github.com/calblueprint/immigration-justice-project/pull/82

CC: @varortz @kyrenetam

### DIFF
--- a/src/app/(auth)/login/page.tsx
+++ b/src/app/(auth)/login/page.tsx
@@ -15,19 +15,23 @@ import { ProfileContext } from '@/utils/ProfileProvider';
 export default function Login() {
   const auth = useAuth();
   const profile = useContext(ProfileContext);
+  if (!auth || !profile) {
+    throw new Error('Page must have auth and profile context defined');
+  }
+
   const [email, setEmail] = useState('');
   const [password, setPassword] = useState('');
   const [errorMessage, setErrorMessage] = useState('');
   const [emailError, setEmailError] = useState('');
   const [passwordError, setPasswordError] = useState('');
-  const [finishLogin, setFinishLogin] = useState(false);
+  const [isLoggingIn, setIsLoggingIn] = useState(false);
+
   const { push } = useRouter();
   const validEmail = (e: string) => e !== '' && isEmail(e);
 
   useEffect(() => {
-    if (!auth) throw new Error('Auth must be defined');
-    if (auth.userId && !finishLogin) push(CONFIG.settings);
-  }, [auth, profile, push]);
+    if (auth.userId && !isLoggingIn) push(CONFIG.settings);
+  }, [auth, profile, push, isLoggingIn]);
 
   const handleSignIn = async () => {
     if (!auth) {
@@ -42,26 +46,29 @@ export default function Login() {
       return;
     }
 
-    const { error } = await auth.signIn(email, password);
+    setIsLoggingIn(true);
+    const { error, data } = await auth.signIn(email, password);
 
     if (error) {
       setErrorMessage(error.message);
       // TODO: use error.status to check if it's an email-specific or password-specific error
       // then, raise the error in the TextInput component.
     } else {
-      profile?.loadProfile();
       setErrorMessage('');
-      setFinishLogin(true);
+      const loadedProfile = await profile.loadProfile(data.user?.id);
+
+      if (!loadedProfile) {
+        push(CONFIG.onboardingHome);
+        return;
+      }
 
       // conditional routing after logging in
-      if (profile?.profileReady) {
-        if (!profile?.profileData) push(CONFIG.onboardingHome);
-        else if (profile.roles.map(r => r.role).includes('ATTORNEY'))
-          push(CONFIG.cases);
-        else if (profile.roles.map(r => r.role).includes('LEGAL_FELLOW'))
-          push(CONFIG.lca);
-        else push(CONFIG.languageSupport);
-      }
+      if (!loadedProfile.profileData) push(CONFIG.onboardingHome);
+      else if (loadedProfile.roles.find(role => role.role === 'ATTORNEY'))
+        push(CONFIG.cases);
+      else if (loadedProfile.roles.find(role => role.role === 'LEGAL_FELLOW'))
+        push(CONFIG.lca);
+      else push(CONFIG.languageSupport);
     }
   };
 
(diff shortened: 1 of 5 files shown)


src/app/(auth)/login/page.tsx
```diff
@@ -42,26 +46,29 @@ export default function Login() {
       return;
     }
 
-    const { error } = await auth.signIn(email, password);
+    setIsLoggingIn(true);
+    const { error, data } = await auth.signIn(email, password);
```
ccatherinetan, line 50: are we using `data` anywhere?

src/app/(auth)/login/page.tsx
```diff
@@ -42,26 +46,29 @@ export default function Login() {
...
-        else if (profile.roles.map(r => r.role).includes('LEGAL_FELLOW'))
-          push(CONFIG.lca);
-        else push(CONFIG.languageSupport);
+      if (!loadedProfile) {
+        push(CONFIG.onboardingHome);
+        return;
       }
+
+      // conditional routing after logging in
+      if (!loadedProfile.profileData) push(CONFIG.onboardingHome);
```
ccatherinetan, line 61: nit: can we combine these two conditions? like: 

`if (!loadedProfile || !loadedProfile.profileData) push(CONFIG.onboardingHome); `

src/utils/ProfileProvider.tsx
```diff
@@ -63,22 +70,36 @@ export default function ProfileProvider({ children }: { children: ReactNode }) {
   const [profileRoles, setProfileRoles] = useState<ProfileRole[]>([]);
   const { userId } = auth;
 
-  const loadProfile = useCallback(async () => {
-    setProfileReady(false);
+  const loadProfile = useCallback(
+    async (id?: string) => {
```
ccatherinetan, line 74: Why did we decide to change loadProfile to take `id` as a parameter, and why didn't we have this before? 
When will `userId` be null/undefined, while `id` is defined?

src/app/(auth)/login/page.tsx
```diff
@@ -15,19 +15,23 @@ import { ProfileContext } from '@/utils/ProfileProvider';
...
 
   useEffect(() => {
-    if (!auth) throw new Error('Auth must be defined');
-    if (auth.userId && !finishLogin) push(CONFIG.settings);
-  }, [auth, profile, push]);
+    if (auth.userId && !isLoggingIn) push(CONFIG.settings);
+  }, [auth, profile, push, isLoggingIn]);
 
   const handleSignIn = async () => {
     if (!auth) {
```
ccatherinetan: I think we can remove this conditional check since we've already checked that auth is defined earlier.

src/app/(auth)/login/page.tsx
```diff
@@ -42,26 +46,29 @@ export default function Login() {
       return;
     }
 
-    const { error } = await auth.signIn(email, password);
+    setIsLoggingIn(true);
```
ccatherinetan, line 44: was there any benefit to moving this up to before the `auth.signIn` function call?

src/app/(auth)/login/page.tsx
```diff
@@ -42,26 +46,29 @@ export default function Login() {
...
 
     if (error) {
       setErrorMessage(error.message);
       // TODO: use error.status to check if it's an email-specific or password-specific error
       // then, raise the error in the TextInput component.
     } else {
-      profile?.loadProfile();
       setErrorMessage('');
-      setFinishLogin(true);
+      const loadedProfile = await profile.loadProfile(data.user?.id);
```
ccatherinetan, line 53: just to clarify, would `data.user` ever be null if the signIn did not error? I'm assuming `data.user?` was added for typing purposes (i.e. since `user` is of type `User | null`)?

ccatherinetan (commented):
looks good, just added a few nits/clarification questions!

src/app/(auth)/login/page.tsx
jinkang-0: we could, but then we'd also have to transform every other check below it to also include `!loadedProfile`

the reason for adding the first `if (!loadedProfile)` check is to just ensure it won't be null - if we join this check with `!loadedProfile.profileData`, then we'd also have to do this for the checks below, since it can be null

both would work, but i think this ultimately reduces the number of boolean checks required

src/utils/ProfileProvider.tsx
jinkang-0: we didn't have this before since the assumption was that user id will always be correct

turns out, for a synchronous execution (i.e. await login, then immediately load profile), the state change happens after load profile, resulting in `loadProfile` detecting user id to be undefined, which gets treated as not logged in, even if log in was successful

therefore, this somewhat hacky solution is simply to allow us to get the userid directly from the return of sign in, which we can then pass onto `loadProfile`.

src/app/(auth)/login/page.tsx
jinkang-0: no clear benefits, but theoretically it should help setup a state change (which is async) before we await another async function, which should help parallelize the flow of execution

src/app/(auth)/login/page.tsx
jinkang-0: i think its just to avoid type errors ye

src/app/(auth)/login/page.tsx
ccatherinetan: will there ever be a case where loadedProfile is truthy, but loadedProfile.profileData) is falsy? 
if not, wouldn't combining the two checks be fine w/o rechecking for every other check?

PR #84, opened 2024-05-16
[fix] typescript complaints of react-select custom styles

jinkang-0:

- Cast all style objects as `CSSObjectWithLabel` to avoid incorrect type inferences

Check if the Typescript error still persists by running:
```bash
npx tsc --noEmit
```

CC: @varortz


jinkang-0:
doesn't seem to affect new PRs, will close for now

PR #85, opened 2024-05-17
[chore] update onboarding flow

jinkang-0:

- Added short description text below State Barred field for Attorneys
- Removed `canContinue` and `setCanContinue` from the onboarding context

image

image

Go in any order

CC: @varortz

### DIFF
--- a/src/components/InputDropdown/index.tsx
+++ b/src/components/InputDropdown/index.tsx
@@ -13,7 +13,7 @@ import Select, {
 import COLORS from '@/styles/colors';
 import { P } from '@/styles/text';
 import { DropdownOption } from '@/types/dropdown';
-import { ErrorText, InputLabel } from '../TextInput/styles';
+import { InputLabel } from '../TextInput/styles';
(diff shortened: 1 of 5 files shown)


src/components/InputDropdown/index.tsx
```diff
@@ -150,7 +150,6 @@ export default function InputDropdown({
         onChange={handleChange}
         onInputChange={handleInputChange}
       />
-      {error && <ErrorText>{error}</ErrorText>}
```
varortz, line 153: are we removing this because we're handling error messages with zod?

jinkang-0: this is mostly because `error` is a boolean, so the error text that shows up displays nothing at all, and only takes up space

jinkang-0: plus, in the usage that we have with `InputDropdown`, they've always been accompanied by `FormMessage` - which shows the error text for us when there is an error, so this becomes redundant

varortz (approved):
Looks good to merge!

PR #86, opened 2024-05-18
[patch] update home page

jinkang-0:

- Added new icons for cases, limited case assignments, and 
- Update hero image + gradient
- Update stats and labels
- Reorientate mission values

image

Go in any order

CC: @varortz, @kyrenetam


src/app/page.tsx
```diff
@@ -18,10 +22,10 @@ type Stat = {
...
 
 const stats: Stat[] = [
-  { number: '180+', label: 'Pro Bono Cases' },
-  { number: '200+', label: 'Attorney Volunteers' },
-  { number: '65+', label: 'Interpreters & Translators' },
-  { number: '50+', label: 'Successful Outcomes' },
+  { number: '180+', label: 'Matched Cases' },
+  { number: '180+', label: 'Attorney' },
+  { number: '65+', label: 'Interpreters' },
+  { number: '45+', label: 'Successful Results' },
```
varortz: can we revert back the stats section to match Ambreen's requested wording + stats: 
image

src/app/page.tsx
```diff
@@ -118,18 +83,56 @@ export default function Home() {
...
       </Styles.StatisticsSection>
       <Styles.HowYouCanHelpContainer style={{ background: COLORS.background }}>
-        <H2 $color={COLORS.blueMid}>
-          HOW <span style={{ color: COLORS.goldMid }}>YOU</span> CAN HELP
+        <H2 $color={COLORS.blueMid} $fontWeight={400}>
+          HOW <strong style={{ color: COLORS.goldMid }}>YOU</strong> CAN HELP
         </H2>
         <Styles.ServicesDiv>
-          {services.map(Service => renderService(Service))}
+          <Styles.ServiceContainer>
```
varortz: is there a reason why we got rid of the `renderService` approach?

src/app/styles.ts
```diff
@@ -106,22 +108,26 @@ export const TextContainer = styled.div`
...
   z-index: 0;
 `;
+
 export const ImageContainer = styled.div`
   position: absolute;
   width: 100%;
   height: 100svh;
   top: 0;
   left: 0;
 `;
```
varortz: we'll probably have to adjust this once we add the navbar right?

src/app/page.tsx
jinkang-0: mostly because it's bad practice to turn a string to HTML tags, as the name implies, this is dangerous (potentially exposes us to XSS attacks) and it's best to not do that if we can help it

src/app/styles.ts
jinkang-0: just checked, ye we'll probably have to adjust it

jinkang-0: cc @kyrenetam

src/app/page.tsx
jinkang-0: but in hindsight, i can bring it back using TSX instead of strings

varortz (approved):
WOWOWOOWOW, this looks so good!! Thanks for updating the home page

PR #88, opened 2024-05-20
[fix] make listing pages fit to screen height

jinkang-0:

Fixed height of listing pages to match that of the screen height, thereby removing the third scrollbar

CC: @varortz, @kyrenetam


varortz (approved):
thank you for this fix!

